PR #5, opened 2018-10-18
bubble not ubble

ctebbe:
:trollface:


prestwich:
keyboard was running out of `b` that day

ctebbe:
> keyboard was running out of `b` that day

no worries mine has some extra `b` today

PR #12, opened 2018-12-05
refactor merkle script to use electrum

prestwich:


prestwich:
@rrybarczyk to review this, please run the script on `master` and then run the script on this branch. Check that the outputs are _exactly_ the same

PR #13, opened 2018-12-13
WIP: Use argparse for cli and add tests

rrybarczyk:


prestwich:
logfile and argparse are kinda overkill here

rrybarczyk:
ok. will remove the log and argparse and keep going with the tests

PR #23, opened 2019-07-29
Relay experiment

prestwich:
Includes #21

This is an experimental relay design. It uses `1 + 1/n` slots per header relayed (n is currently 4), and 2 slots to externalize useful information (best chain tip and best shared ancestor of latest reorg).

The core idea behind the relay is to minimize storage costs by increasing `calldata` costs. Rather than storing headers, the relay stores the `hashPrevBlock` field of each header and the height of every `n`th header. Should the relay need to reference information in old headers (like the difficulty), the header data is passed to the relay again, and validated against known `hashPrevBlock` links. This allows the relay to check that newly submitted blocks are valid extensions of existing blocks, without storing all past header information.

As opposed to other relays, we separate the function of the relay into two categories: "learning about new blocks" and "following the best chain tip." Users may add new blocks in groups of at least 5 by calling `addHeaders` If the block slice includes a difficulty retarget, users are required to call `addHeadersWithRetarget`, which performs additional validation. The relay does not update its tip unless it is specifically requested to do so by a user. The user must call `markNewHeaviest` with the new heaviest, the old heaviest header, and the digest of their most recent common ancestor (which may be the old heaviest header. 

As part of the process, the relay externalizes the most recent common ancestor, which is to say, the heaviest header that both old and new heaviest tip confirm. This is a metric of "subjective finality" for that block. During normal operation without reorgs it lags behind the tip by 5 blocks. During reorgs, it is the shared base of the competing branches (and as such may move backwards!). This indicates that competing sets of miners both viewed it as subjectively finalized. As such, it is a reasonable source of finalization information for relay-consuming smart contracts.

This model provides large gas savings compared to previous relay designs (TODO: benchmarking). It also gets especially attractive if EIP2028 activates, reducing calldata gas costs.


pyskell:
I'd say move up some of your `require()`s where possible to save on gas/cpu cycles.
https://github.com/summa-tx/bitcoin-spv/blob/9b479dd898061b78db58c0d69e365b07bef4a683/contracts/Relay.sol#L94

Also this message isn't really informative :wink: https://github.com/summa-tx/bitcoin-spv/blob/9b479dd898061b78db58c0d69e365b07bef4a683/contracts/Relay.sol#L70

pyskell:
Why not just require this?
https://github.com/summa-tx/bitcoin-spv/blob/9b479dd898061b78db58c0d69e365b07bef4a683/contracts/Relay.sol#L126

Also wish I could give better feedback but I'm not knowledgeable about spvs.

prestwich:
> I'd say move up some of your require()s where possible to save on gas/cpu cycles.

We did do this, but there are a few places where testing becomes extremely difficult if you rearrange requires. I think for the remaining pieces, the gas savings would be quite small

> Also this message isn't really informative :wink:

I think it's extremely descriptive. It tells the user exactly how to fix their error :stuck_out_tongue_closed_eyes: 

> Why not just require this?

I'm going to remove the comment, as it's not as relevant as it was in an earlier design stage. One of the next pushes from this will be to generalize the relay for other chains which might have different retarget systems (or header lengths, etc.)

Of course, relays will (by necessity) always be more expensive on chains with shorter periods or more complex retarget systems

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+        require(_headers.length / 80 >= 5, "Must supply at least 5 headers");
+
+        /*
+        NB:
+        1. check that the header has sufficient work
+        2. check that headers are in a coherent chain (no retargets, hash links good)
+        3. Store the block connection
+        4. Store the height
+        */
+        for (uint256 i = 0; i < _headers.length / 80; i = i.add(1)) {
```
gakonst: nit: move `_headers.length / 80` in a variable outside of the loop. The EVM SLOADs array lengths per iteration as it cannot know if the length has changed in the loop. Check this in remix:


```Solidity

pragma solidity ^0.5.1;

contract C {
    uint256[] x;
    
    constructor() public {
        for (uint256 i = 0 ; i < 50; i++ ) {
            x.push(i);
        } 
    }
    
    function cheaper() public {
        uint256 y = 0;
        uint256 len = x.length;
        for (uint256 i = 0 ; i < len; i++ ) {
            y+=1;
        }
    }
    
    function costly() public {
        uint256 y = 0;
        for (uint256 i = 0 ; i < x.length; i++ ) {
            y+=1;
        }
    }
}
```

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+            _header = _headers.slice(i.mul(80), 80);
+            _height = _anchorHeight.add(i).add(1);
```
gakonst: Is this double add optimized by the compiler? Might be better to do `.add(i+1)`

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+            _currentDigest = _header.hash256();
+
+            /*
+            NB:
+            if the block is already authenticated, we don't need to a work check
+            Or write it anything to state.
```
gakonst: typo in comment

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+            This saves gas
+            */
+            if (previousBlock[_currentDigest] == bytes32(0)) {
+                require(
+                    abi.encodePacked(_currentDigest).reverseEndianness().bytesToUint() <= _target,
+                    "Header work is insufficient");
+                previousBlock[_currentDigest] = _previousDigest;
+                if (_height % 4 == 0) {
```
gakonst: Can't you store just the Nth block when providing with a header chain of length N? How is 4 different from any other parameter?

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+                    /*
+                    NB: We store the height only every 4th header to save gas
+                    */
+                    blockHeight[_currentDigest] = _height;
+                }
+            }
+
+            /* NB: we do still need to make chain level checks tho */
+            require(_header.extractTarget() == _target, "Target changed unexpectedly");
+            require(_header.validateHeaderPrevHash(_previousDigest), "Headers not a consistent chain");
```
gakonst:
```suggestion
            require(_header.validateHeaderPrevHash(_previousDigest), "Headers do not form a consistent chain");
```

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /// @notice                 Adds headers to storage, performs additional validation of retarget
+    /// @dev                    Checks the retarget, the heights, and the linkage
+    /// @param  _oldPeriodStart The first header in the difficulty period being closed
+    /// @param  _oldPeriodEnd   The last header in the difficulty period being closed
+    /// @param  _headers        A tightly-packed list of 80-byte Bitcoin headers
+    /// @return                 True if successfully written, error otherwise
+    function addHeadersWithRetarget(
+        bytes calldata _oldPeriodStart,
```
gakonst: nit: 
```suggestion
        bytes calldata _oldPeriodStartHeader,
```

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+
+        /* NB: This comparison looks weird because header nBits encoding truncates targes */
+        uint256 _actualTarget = _newPeriodStart.extractTarget();
+        uint256 _expectedTarget = BTCUtils.retargetAlgorithm(
+            _oldPeriodStart.extractTarget(),
+            _oldPeriodStart.extractTimestamp(),
+            _oldPeriodEnd.extractTimestamp()
+        );
+        require(
+            (_actualTarget & _expectedTarget) == _actualTarget,
```
gakonst: What is this check for?

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /// @notice         Finds the height of a header by its digest
+    /// @dev            Will fail if the header is unknown
+    /// @param _digest  The header digest to search for
+    /// @return         The height of the header
+    function _findHeight(bytes32 _digest) internal view returns (uint256) {
+        uint256 _height = 0;
+        bytes32 _current = _digest;
+        for (uint256 i = 0; i < 5; i = i.add(1)) {
```
gakonst: Should this only search for 5 blocks after the current one?

contracts/Relay.sol
```diff
@@ -0,0 +1,431 @@
...
+
+
+    /// @notice                   Gives a starting point for the relay
+    /// @dev                      We don't check this AT ALL really. Don't use relays with bad genesis
+    /// @param  _genesisHeader    The starting header
+    /// @param  _height           The starting height
+    constructor(bytes memory _genesisHeader, uint256 _height, bytes32 _periodStart) public {
+        require(_genesisHeader.length == 80, "Stop being dumb");
+        bytes32 _genesisDigest = _genesisHeader.hash256();
+        relayGenesis = _genesisDigest;
```
gakonst: nit: This doesn't seem useful at all in the context of stateless spv, so consider removing it

contracts/Relay.sol
prestwich: Given that this is a memory array, presumably it `MLOAD`s it?

contracts/Relay.sol
prestwich: height is stored every 4, so if we traverse 5 and find nothing, it doesn't exist.

contracts/Relay.sol
prestwich: Bitcoin difficulty is generated at full-bitlength, but only the 1.5 most-significant non-zero bytes are encoded into headers. So when doing the calculation, `_expectedTarget` uses all bits. We have to `AND` mask it with the blockheader-encoded target (`_actualTarget`) before making the equality comparison.

Basically we check if the observed encoded difficulty (pulled from the header) matches the expected encoded difficulty (calculated from the previous encoded difficulty and the timestamps).

contracts/Relay.sol
prestwich: I don't believe it's optimized. So this repeats the SafeMath overflow check. Marginal gas savings at (essentially) no risk from changing it

prestwich: tecccchhhhnically the check is not redundant. But it should be impossible to overflow here eitherway

contracts/Relay.sol
prestwich: It's marginally useful for consumers to be able to decide if the relay started from an okay block

contracts/Relay.sol
prestwich: I'm really on the fence about allowing this to be parameterized. Ideally, this contract is a chain-wide singleton. 4 seems like a reasonable balance between saving money on writes and keeping `findHeight` calls cheap

4 means 5,000 gas per header (amortized)
and lookups cost >=600 (3 `SLOAD`)

8 would mean 2,500 gas per header (amortized)
and lookups costing  >=1,400 (7 `SLOAD`)

Considering we expect to do >1 lookup per 4 headers, (1 for extending the chain, and some amount for marking new tips), it seems like 8 is too sparse

contracts/Relay.sol
gakonst: OK so make `4` a constant at the top of the file and make the loop end at `constant+1`, the 5 looks arbitrary to a reader otherwise

prestwich:
This has been moved into its own repo

PR #25, opened 2019-08-05
WIP: Convert BTCUtils and ValidateSPV to Javascript

ErinHales:


contracts-js/ValidateSPV.js
```diff
@@ -0,0 +1,315 @@
...
+
+  ERR_BAD_LENGTH: 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff,
+  ERR_INVALID_CHAIN: 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe,
+  ERR_LOW_WORK: 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd,
+
+//     function getErrBadLength() internal pure returns (uint256) {
+//         return ERR_BAD_LENGTH;
+//     }
+
+  getErrBadLength: () => {
```
prestwich: Can remove these getters. They're required in Solidity, but not so much in JS

contracts-js/BTCUtils.js
```diff
@@ -0,0 +1,1012 @@
...
+
+//         return _mantissa * (256 ** _exponent);
+//     }
+
+  /// @notice          Extracts the target from a block header
+  /// @dev             Target is a 256 bit number encoded as a 3-byte mantissa and 1 byte exponent
+  /// @param _header   The header
+  /// @return          The target threshold
+  extractTarget: (header) => {
+    let d_header = utils.deserializeHex(header)
```
prestwich: The header should already be deserialized. Which is to say, header should be a `Uint8Array`

utils/utils.js
```diff
@@ -0,0 +1,73 @@
...
+    if (a.byteLength !== b.byteLength) return false;
+    return a.every((val, i) => val === b[i]);
+  },
+
+  /// @notice          Converts big-endian bytes to a uint
+  /// @dev             Traverses the byte array and sums the bytes
+  /// @param _b        The big-endian bytes-encoded integer
+  /// @return          The integer representation
+  bytesToUint: (bytesString) => {
+    // const newString = bytesString.slice(2) // This copies the array, minus the '0x' prefix.
```
prestwich: This should also take `Uint8Array` instead of a string as argument

utils/utils.js
```diff
@@ -0,0 +1,73 @@
...
+
+    return `0x${hexStr.toLowerCase()}`
+  },
+
+  deserializeHex: (hexStr) => {
+    if (!hexStr) {
+      return new Uint8Array()
+    }
+
+<<<<<<< HEAD
```
prestwich: Merge conflict :D

contracts-js/BTCUtils.js
```diff
@@ -0,0 +1,1012 @@
...
+
+    // Hacky way of reversing endianness of a partial serialized number
+    let m = header.slice(72, 75).split().reverse().join('')
```
prestwich: This will get characters 72-74 of a hex string. We want bytes 72-74 of a `Uint8Array`.

contracts-js/BTCUtils.js
```diff
@@ -0,0 +1,1012 @@
...
+    let hex_m = `0x${m}`
+
+    let e = d_header[75] - 3
```
prestwich: `let e = BigInt(d_header[75] - 3)`

contracts-js/BTCUtils.js
```diff
@@ -0,0 +1,1012 @@
...
+    let mantissa = utils.bytesToUint(hex_m)
+
+    console.log('mantissa: ', mantissa)
+    // console.log('e: ', e)
+    // console.log('m: ', m)
+    // let exponent = BigInt(256 ** (e-3))
+    // console.log('exponent: ', exponent)
+    // let exponent = 256 ** (e - 3)
+
+    return mantissa * exponent
```
prestwich:
```
let exponent = (e - 3n)
return mantissa * 256n ** exponent
```

contracts-js/BTCUtils.js
dominiquejane: it's not. console.log(header) gives `0x0100000055bd840a78798ad0da853f68974f3d183e2bd1db6a842c1feecf222a00000000ff104ccb05421ab93e63f8c3ce5c2c2e9dbb37de2764b3a3175c8166562cac7d51b96a49ffff001d283e9e70`

contracts-js/BTCUtils.js
dominiquejane: Made changes to previous comments.
I guess node just doesn't understand BigInt. It throws a TypeError: `Do not know how to serialize a BigInt`. This occurs when something is trying to use JSON.stringify on the result.

utils/utils.js
dominiquejane: removed string manipulation functions.

contracts-js/ValidateSPV.js
ErinHales: Done

test/BTCUtilsJs.test.js
```diff
@@ -0,0 +1,442 @@
...
+    assert.equal(res, '0x40420f0000000000220020aedad4518f56379ef6f1f52f2e0fed64608006b3ccaff2253d847ddc90c91922');
+  });
+
+  it('extracts a root from a header', async () => {
+    const res = await BTCUtilsJs.extractMerkleRootBE(HEADER_170);
+    assert.equal(res, '0x7dac2c5666815c17a3b36427de37bb9d2e2c5ccec3f8633eb91a4205cb4c10ff');
+  });
+
+  it('extracts the target from a header', async () => {
+    const res = await BTCUtilsJs.extractTarget(HEADER_170);
```
prestwich: @dominiquejane if this is the test that's running, you need to call `deserializeHex` here before passing the header in

utils/utils.js
ErinHales: I just pulled the code a few minutes ago and resolved this

contracts-js/BTCUtils.js
```diff
@@ -0,0 +1,1012 @@
...
+  //     using BytesLib for bytes;
+//     using SafeMath for uint256;
+
+//     // The target at minimum Difficulty. Also the target of the genesis block
+//     uint256 public constant DIFF1_TARGET = 0xffff0000000000000000000000000000000000000000000000000000;
+
+//     uint256 public constant RETARGET_PERIOD = 2 * 7 * 24 * 60 * 60;  // 2 weeks in seconds
+//     uint256 public constant RETARGET_PERIOD_BLOCKS = 2016;  // 2 weeks in blocks
+  RETARGET_PERIOD: 1209600,
+  RETARGET_PERIOD_BLOCKS: 2016,
```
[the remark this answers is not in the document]
dominiquejane: done

contracts-js/BTCUtils.js
prestwich: > This occurs when something is trying to use JSON.stringify on the result.

Ya it's not ideal. `JSON.stringify` will probably be broken for them for a while

test/BTCUtilsJs.test.js
dominiquejane: ok. Other tests also use this header. Do they all need to be deserialized first?

prestwich: Yes. It might be wise to simply deserialize the header where we declare it a constant in the header file

prestwich: in the test file*

test/BTCUtilsJs.test.js
```diff
@@ -0,0 +1,442 @@
+/* global artifacts contract before it assert */
+// const BN = require('bn.js');
+const constants = require('./constants');
+
+const assert = require('chai').assert
+
+const HEADER_170 = '0x0100000055bd840a78798ad0da853f68974f3d183e2bd1db6a842c1feecf222a00000000ff104ccb05421ab93e63f8c3ce5c2c2e9dbb37de2764b3a3175c8166562cac7d51b96a49ffff001d283e9e70';
```
prestwich: Deserialize here?  (and for the rest of the `const`s on the next several lines)

test/BTCUtilsJs.test.js
dominiquejane: k

test/BTCUtilsJs.test.js
dominiquejane: done

contracts-js/BTCUtils.js
dominiquejane: k

utils/utils.js
dominiquejane: @prestwich can you check this function is doing what it's supposed to do?

test/BTCUtilsJs.test.js
```diff
@@ -0,0 +1,449 @@
...
+    /* eslint-disable no-await-in-loop */
+    let firstTimestamp;
+    let secondTimestamp;
+    let previousTarget;
+    let expectedNewTarget;
+    let res;
+    for (let i = 0; i < constants.RETARGET_TUPLES.length; i += 1) {
+      firstTimestamp = constants.RETARGET_TUPLES[i][0].timestamp;
+      secondTimestamp = constants.RETARGET_TUPLES[i][1].timestamp;
+      previousTarget = await BTCUtilsJs.extractTarget.call(`0x${constants.RETARGET_TUPLES[i][1].hex}`);
```
dominiquejane: @prestwich All the tests that use `RETARGET_TUPLES` are using them as serialized values. How do you want us to address these since we're writing the functions to accept u8a?

prestwich: Deserialize them wherever they're passed into a test. In this case it'd be something like:
```
firstTimestamp = BigInt(constants.RETARGET_TUPLES[i][0].timestamp);
secondTimestamp = BigInt(constants.RETARGET_TUPLES[i][1].timestamp);
previousTarget = BTCUtilsJs.extractTarget.call(utils.deserializeHex(constants.RETARGET_TUPLES[i][1].hex));

File: utils/utils.js (hunk 3 above)
Comment (prestwich) [reply utils/utils.js]: It should accept a u8a instead of a string.
```
(buf) => {
    let total = BigInt(0)
    for (var i = 0; i < buf.length; i++) {
      total += BigInt(buf[i]) << (BigInt(buf.length - i - 1) * BigInt(8))
    }
    return total
}```

contracts-js/BTCUtils.js
```diff
@@ -0,0 +1,1031 @@
+// /** @title BitcoinSPV */
+// /** @author Summa (https://summa.one) */
+
+// import {BytesLib} from "./BytesLib.sol";
+// import {SafeMath} from "./SafeMath.sol";
+
+// sha256 --> Hold off on this for now
+// ripemd160 --> Hold off on this for now
+// keccack256 --> Can just drop this and do an equality test instead, James will find equality test to use
```
tynes: Here is a `safeEqual` function that compares buffers in constant time
https://github.com/bcoin-org/bcrypto/blob/91bfabae43135ead6c2b2e43833f3a2325ebd737/lib/safe.js#L46

prestwich: @ErinHales @dominiquejane 

constant-time implementations are amazing in off-chain cryptography. They make it so that you can't learn any secrets by timing how long a computation takes. But unfortunately, they're more expensive than most algorithms, so we probably won't use them for things that run on-chain (where there are no secrets anyway)

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @param {number}        index The 0-indexed location of the output to extract
+   * @returns {Uint8Array}  The specified output
+   */
+  extractOutputAtIndex: (vout, index) => {
+    let len;
+    let remaining;
+    let offset = 1n;
+
+    for (let i = 0; i <= index; i++) {
+      remaining = utils.safeSlice(vout, offset, vout.length - 1);
```
prestwich: Is this `- 1` supposed to be here?

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @param {index}         index The 0-indexed location of the input to extract
+   * @returns {Uint8Array}  The input as a u8a
+   */
+  extractInputAtIndex: (vinArr, index) => {
+    let len = 0;
+    let remaining = 0;
+    let offset = 1n;
+
+    for (let i = 0; i <= index; i++) {
+      remaining = utils.safeSlice(vinArr, offset, vinArr.length - 1);
```
prestwich: Is this `- 1` supposed to be here?  Doesn't this make `remaining` always miss 1 byte off the end of the array?

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @notice                Extracts the output at a given index in the TxIns vector
+   * @dev                   Iterates over the vout. If you need to extract multiple, write a custom function
+   * @param {Uint8Array}    vout The _vout to extract from
+   * @param {number}        index The 0-indexed location of the output to extract
+   * @returns {Uint8Array}  The specified output
+   */
+  extractOutputAtIndex: (vout, index) => {
+    let len;
+    let remaining;
+    let offset = 1n;
```
prestwich: Can we move `let len` and `let remaining` into the loop

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   */
+  extractValueLE: (output) => {
+    return utils.safeSlice(output, 0, 8);
+  },
+
+  /**
+   * @notice                Extracts the value from the output in a tx
+   * @dev                   Value is an 8-byte little-endian number
+   * @param {Uint8Array}    output The output
+   * @returns {Uint8Array}  The output value
```
prestwich: This returns a BigInt

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+
+  /**
+   * @notice                Extracts the data from an op return output
+   * @dev                   Returns empty Uint8Array if no data or not an op return
+   * @param {Uint8Array}    output The output
+   * @returns {Uint8Array}  Any data contained in the opreturn output, null if not an op return
+   */
+  extractOpReturnData: (output) => {
+    if (!utils.typedArraysAreEqual(utils.safeSlice(output, 9, 10), new Uint8Array([106]))) {
+      return null;
```
prestwich: let's make this `throw new Error('Not an OP_RETURN output')`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  },
+
+  /**
+   * @notice                Extracts the data from an op return output
+   * @dev                   Returns empty Uint8Array if no data or not an op return
+   * @param {Uint8Array}    output The output
+   * @returns {Uint8Array}  Any data contained in the opreturn output, null if not an op return
+   */
+  extractOpReturnData: (output) => {
+    if (!utils.typedArraysAreEqual(utils.safeSlice(output, 9, 10), new Uint8Array([106]))) {
```
prestwich: there's a simpler way of doing this `output[9] !== 106`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+      return null;
+    }
+    let dataLen = utils.safeSlice(output, 10, 11);
+    return utils.safeSlice(output, 11, 11 + Number(utils.bytesToUint(dataLen)));
```
prestwich: there's a simpler way to do this:
```
let dataLen = output[10];
return utils.safeSlice(output, 11, 11 + dataLen);
```

Remember that u8as are arrays of `Numbers` between 0 and 255. So when we want a single element, we can just index it and treat it as a Number

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @dev                   Determines type by the length prefix and validates format
+   * @param {Uint8Array}    output The output
+   * @returns {Uint8Array}  The hash committed to by the pk_script, or null for errors
+   */
+  extractHash: (output) => {
+    if (utils.safeSlice(output, 9, 10)[0] == 0) {
+      let len = module.exports.extractOutputScriptLen(output)[0] - 2;
+      // Check for maliciously formatted witness outputs
+      if (utils.safeSlice(output, 10, 11)[0] != len) {
+        return null;
```
prestwich: `throw new Error('Malformatted witness output')`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+      }
+      return utils.safeSlice(output, 11, 11 + len);
+    } else {
+      let tag = utils.safeSlice(output, 8, 11);
+      // p2pkh
+      if (utils.typedArraysAreEqual(tag, utils.deserializeHex('0x1976a9'))) {
+        // Check for maliciously formatted p2pkh
+        if (utils.safeSlice(output, 11, 12)[0] != 0x14 || !utils.typedArraysAreEqual(utils.safeSlice(output, output.length - 2, output.length), utils.deserializeHex('0x88ac'))) {
```
prestwich: Clean up: 
```
let lastTwoBytes = utils.safeSlice(output, output.length - 2, output.length);
if (output[11] != 0x14 || !utils.typedArraysAreEqual(lastTwoBytes, new Uint8Array([0x88, 0xac])))
```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+      let len = module.exports.extractOutputScriptLen(output)[0] - 2;
+      // Check for maliciously formatted witness outputs
+      if (utils.safeSlice(output, 10, 11)[0] != len) {
+        return null;
+      }
+      return utils.safeSlice(output, 11, 11 + len);
+    } else {
+      let tag = utils.safeSlice(output, 8, 11);
+      // p2pkh
+      if (utils.typedArraysAreEqual(tag, utils.deserializeHex('0x1976a9'))) {
```
prestwich: clean up:

```
if (utils.typedArraysAreEqual(tag, new Uint8Array([0x19, 0x76, 0xa9]))) {
```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+        // Check for maliciously formatted p2pkh
+        if (utils.safeSlice(output, 11, 12)[0] != 0x14 || !utils.typedArraysAreEqual(utils.safeSlice(output, output.length - 2, output.length), utils.deserializeHex('0x88ac'))) {
+          return null;
```
prestwich: `throw new Error('Malformatted P2PKH output')`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+        }
+        return utils.safeSlice(output, 12, 32);
+      //p2sh
+      } else if (utils.typedArraysAreEqual(tag, utils.deserializeHex('0x17a914'))) {
```
prestwich: ```if (utils.typedArraysAreEqual(tag, new Uint8Array([0x17, 0xa9, 0x14]))) {```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+        // Check for maliciously formatted p2sh
+        if (utils.safeSlice(output, output.length - 1, output.length)[0] != 0x87) {
```
prestwich:
```
output[output.length - 1] != 87
```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+          return null;
```
prestwich: `throw new Error('Malformatted P2SH output')`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+        }
+        return utils.safeSlice(output, 11, 31);
+      }
+    }
+    return null;  /* NB: will trigger on OPRETURN and non-standard that don't overrun */
```
prestwich: `throw new Error('Nonstandard, OP_RETURN or malformatted output')`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  /**
+   * @notice                  Checks that the vin passed up is properly formatted
+   * @dev                     Consider a vin with a valid vout in its scriptsig
+   * @param {Uint8Array}      vin Raw bytes length-prefixed input vector
+   * @returns {Boolean}       True if it represents a validly formatted vin
+   */
+  validateVin: (vin) => {
+    let offset = 1n;
+    let vLength = BigInt(vin.length);
+    nIns = utils.safeSlice(vin, 0, 1)[0];
```
prestwich: nIns = vin[0]

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+
+    // Returns false if we're not exactly at the end
+    return offset == vLength;
+  },
+
+  /**
+   * @notice                Checks that the vout passed up is properly formatted
+   * @dev                   Consider a vin with a valid vout in its scriptsig
+   * @param {Uint8Array}    vout Raw bytes length-prefixed output vector
+   * @returns {Boolean}     True if it represents a validly formatted bout
```
prestwich: `bout`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   */
+  validateVout: (vout) => {
+    offset = 1n;
+    let vLength = BigInt(vout.length);
+    nOuts = utils.safeSlice(vout, 0, 1)[0];
```
prestwich: `nOuts = vout[0]`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @notice                Calculate difficulty from the difficulty 1 target and current target
+   * @dev                   Difficulty 1 is 0x1d00ffff on mainnet and testnet
+   * @dev                   Difficulty 1 is a 256 bit number encoded as a 3-byte mantissa and 1 byte exponent
+   * @param {BigInt/number} target The current target
+   * @returns {BigInt}      The block difficulty (bdiff)
+   */
+  calculateDifficulty: (target) => {
+    if (typeof target !== 'bigint') {
+      let bigTarget = BigInt(target);
+      return module.exports.DIFF1_TARGET / bigTarget;
```
prestwich: simpler:

```
    if (typeof target !== 'bigint') {
      target = BigInt(target);
    }
return module.exports.DIFF1_TARGET / target;
```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  },
+
+  /**
+   * @notice                Extracts the timestamp from a block header
+   * @dev                   Time is not 100% reliable
+   * @param {Uint8Array}    header The header
+   * @returns {BigInt}      The timestamp (uint)
+   */
+  extractTimestamp: (header) => {
+    return utils.bytesToUint(module.exports.reverseEndianness(module.exports.extractTimestampLE(header)));
```
prestwich: let's break this up into a few lines

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @returns {Uint8Array}  The double-sha256 of the concatenated hashes
+   */
+  hash256MerkleStep: (a, b) => {
+    return module.exports.hash256(utils.concatUint8Arrays([a, b]));
+  },
+
+  /**
+   * @notice                Verifies a Bitcoin-style merkle tree
+   * @dev                   Leaves are 1-indexed.
+   * @param {Uin8Array}     proof The proof. Tightly packed LE sha256 hashes. The last hash is the root
```
prestwich: missing a `t` in u8a

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+    if (proofLength === 32) {
+      return true;
+    }
+
+    // Should never occur
+    if (proofLength === 64) {
+      return false;
+    }
+
+    let idx = BigInt(index);
```
prestwich: does index need to be a BigInt?

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+    let root = utils.safeSlice(proof, (proofLength - 32), proofLength);
+    let current = utils.safeSlice(proof, 0, 32);
+
+    for (let i = 1; i < ((proofLength / 32) - 1); i++) {
```
prestwich: let's add a line here:
```
let next = utils.safeSlice(proof, (i * 32), ((i * 32) + 32))
```

then we can make the following block:

```
if (idx % 2n === 1n) {
        current = module.exports.hash256MerkleStep(next, current);
      } else {
        current = module.exports.hash256MerkleStep(current, next);
      }
```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+   * @dev                   Implements the Bitcoin algorithm precisely
+   * @param {BigInt}        previousTarget The target of the previous period
+   * @param {number}        firstTimestamp The timestamp of the first block in the difficulty period
+   * @param {number}        secondTimestamp The timestamp of the last block in the difficulty period
+   * @returns {BigInt}      The new period's target threshold
+   */
+  retargetAlgorithm: (previousTarget, firstTimestamp, secondTimestamp) => {
+    let elapsedTime = BigInt(secondTimestamp - firstTimestamp);
+    const rp = module.exports.RETARGET_PERIOD;
+    const div = rp / 4n;
```
prestwich: call this `lowerBound`

prestwich: or `floor`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+    const mult = rp * 4n;
```
prestwich: call this `upperBound` or `ceiling`

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+    // Normalize ratio to factor of 4 if very long or very short
+    if (elapsedTime < div) {
+      elapsedTime = div;
+    }
+    if (elapsedTime > mult) {
+      elapsedTime = mult;
+    }
+
+    /*
+      NB: high targets e.g. ffff0020 can cause overflows here
```
prestwich: overflows in JS aren't a thing. We can remove this comment and also the 65536

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+          so we divide it by 256**2, then multiply by 256**2 later
+          we know the target is evenly divisible by 256**2, so this isn't an issue
+    */
+
+    let adjusted = (previousTarget / 65536n) * elapsedTime;
+    return (adjusted / module.exports.RETARGET_PERIOD) * 65536n;
```
prestwich:
```
    return previousTarget * elapsedTime / module.exports.RETARGET_PERIOD;
```

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  },
+
+  /**
+   * @notice                 Get the last _num bytes from a byte array
+   * @dev                    The byte array to slice
+   * @param {Uint8Array}     uint8Arr The big-endian array-encoded integer
+   * @returns {BigInt}       The integer representation
+   */
+  lastBytes: (arr, num) => {
+    if (num >= arr.length) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  hash160: (preImage) => {
+    return utils.ripemd160(utils.sha256(preImage));
+  },
+
+/**
+ * @notice                  Implements bitcoin's hash256 (double sha2)
+ * @param {Uint8Array}      preImage The pre-image
+ * @returns {Uint8Array}    The digest
+ */
+  hash256: (b) => {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  /**
+   * @notice                Extracts the nth input from the vin (0-indexed)
+   * @dev                   Iterates over the vin. If you need to extract several, write a custom function
+   * @param {Uint8Array}    vinArr The vin as a tightly-packed uint8array
+   * @param {index}         index The 0-indexed location of the input to extract
+   * @returns {Uint8Array}  The input as a u8a
+   */
+  extractInputAtIndex: (vinArr, index) => {
+    let len = 0;
+    let remaining = 0;
```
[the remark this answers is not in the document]
ErinHales: `len` needs to be outside the for loop so it can be accessed in the return statement. `remaining` can be moved inside the loop.

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  },
+
+  /**
+   * @notice                Determines whether an input is legacy
+   * @dev                   False if no scriptSig, otherwise True
+   * @param {Uint8Array}    input The input
+   * @returns {boolean}     True for legacy, False for witness
+   */
+  isLegacyInput: (input) => {
+    return !utils.typedArraysAreEqual(utils.safeSlice(input, 36, 37), new Uint8Array([0]));
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
...
+  },
+
+  /**
+   * @notice                Determines the length of an input from its scriptsig
+   * @dev                   36 for outpoint, 1 for scriptsig length, 4 for sequence
+   * @param {Uint8Array}    arr The input as a u8a
+   * @returns {BigInt}      The length of the input in bytes
+   */
+  determineInputLength: (arr) => {
+    let res = module.exports.extractScriptSigLen(arr);
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: smart, fixed

js/lib/BTCUtils.js
ErinHales: fixed, used `lowerBound`

js/lib/BTCUtils.js
ErinHales: fixed, used `upperBound`

js/lib/BTCUtils.js
ErinHales: So ```let adjusted = (previousTarget / 65536n) * elapsedTime;
    return (adjusted / module.exports.RETARGET_PERIOD) * 65536n;``` becomes ```let adjusted = (previousTarget) * elapsedTime;
    return (adjusted / module.exports.RETARGET_PERIOD);```?

ErinHales: Never mind, next comment answers that.  Fixed.

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+   * @notice                Validates a tx inclusion in the block
+   * @param {Uint8Array}    txid The txid (LE)
+   * @param {Uint8Array}    merkleRoot The merkle root (as in the block header)
+   * @param {Uint8Array}    intermediateNodes The proof's intermediate nodes (digests between leaf and root)
+   * @param {number}        index The leaf's index in the tree (0-indexed)
+   * @returns {boolean}     true if fully valid, false otherwise
+   */
+  prove: (txid, merkleRoot, intermediateNodes, index) => {
+    // Shortcut the empty-block case
+    if (utils.typedArraysAreEqual(txid, merkleRoot) && index === 0 && intermediateNodes.length == 0) {
```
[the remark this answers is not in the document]
ErinHales: added triple equals

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * @notice                Parses a tx input from raw input bytes
+   * @dev                   Supports Legacy and Witness inputs
+   * @param {Uint8Array}    input bytes tx input
+   * @returns {object}      Tx input, sequence number, tx hash, and index
+   */
+  parseInput: (input) => {
+    // NB: If the scriptsig is exactly 00, we are witness.
+    // Otherwise we are compatibility
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+    let sequence;
+    let witnessTag;
+    let inputType;
+
+    if (!utils.typedArraysAreEqual(utils.safeSlice(input, 36, 37), new Uint8Array([0]))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+   * @notice                Parses a block header struct from a bytestring
+   * @dev                   Block headers are always 80 bytes, see Bitcoin docs
+   * @param {Uint8Array}    header Header
+   * @returns {object}      Header digest, version, previous block header hash, merkle root, timestamp, target, nonce
+   */
+  parseHeader: (header) => {
+    // If header has an invalid length, bubble up error
+    const EMPTY = utils.deserializeHex('0x0000000000000000000000000000000000000000000000000000000000000000');
+    if (header.length != 80) {
+      return { digest: EMPTY, version: 0n, prevHash: EMPTY, merkleRoot: EMPTY, timestamp: 0n, target: 0n, nonce: 0n };
```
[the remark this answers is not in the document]
ErinHales: fixed

js/utils/sha256.js
```diff
@@ -0,0 +1,518 @@
...
+      } else {
+        if (message === null || message === undefined) {
+          throw new Error(ERROR);
+        } else if (message.constructor === ArrayBuffer) {
+          message = new Uint8Array(message);
+        }
+      }
+      if (Array.isArray(message) || ArrayBuffer.isView(message) ||
+        message.constructor === Buffer) {
+        return crypto.createHash(algorithm).update(new Buffer(message)).digest('hex');
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/BTCUtils.js
```diff
@@ -0,0 +1,659 @@
+/** @title BitcoinSPV */
+/** @author Summa (https://summa.one) */
```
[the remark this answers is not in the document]
ErinHales: Here's something. I don't think we need the `last modified` bit, but we can copy the format. https://marketplace.visualstudio.com/items?itemName=arjunkomath.js-file-header

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
+// /** @title ValidateSPV*/
```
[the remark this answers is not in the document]
ErinHales: Replaced with js docs

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+    // Otherwise we are compatibility
+    let sequence;
+    let witnessTag;
+    let inputType;
+
+    if (!utils.typedArraysAreEqual(utils.safeSlice(input, 36, 37), new Uint8Array([0]))) {
+      sequence = btcUtils.extractSequenceLegacy(input);
+      witnessTag = utils.safeSlice(input, 36, 39);
+
+      if (utils.typedArraysAreEqual(witnessTag, utils.deserializeHex('220020')) || utils.typedArraysAreEqual(witnessTag, utils.deserializeHex('160014'))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+  /**
+   * @notice                Parses a tx output from raw output bytes
+   * @dev                   Differentiates by output script prefix, handles legacy and witness
+   * @param {Uint8Array}    output bytes tx output
+   * @returns {object}      Tx output value, output type, payload
+   */
+  parseOutput: (output) => {
+    let value = btcUtils.extractValue(output);
+
+    if (utils.typedArraysAreEqual(utils.safeSlice(output, 9, 10), new Uint8Array([106]))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "bitcoin-spv-js",
+    "author": "Erin Hales, Dominique Liau, James Prestwich",
```
tynes: More of a nit to help out automation around package management and introspection, the `author` field here doesn't fit the schema defined by npm - https://docs.npmjs.com/files/package.json#people-fields-author-contributors

js/package.json
```diff
@@ -0,0 +1,19 @@
+{
```
tynes: Since you depend on native bigint, you should have `engineStrict` set to greater than or equal to node v10 because that was the version they were added. Should also add that to the readme because people don't always look at the `engine` field in the `package.json`

js/utils/utils.js
```diff
@@ -0,0 +1,116 @@
...
+    return module.exports.deserializeHex(shaLib(buf));
+  },
+
+  ripemd160: (buf) => {
+    return ripemd160Lib.default(buf);
+  },
+
+  typedArraysAreEqual: (a, b) => {
+    if (a.byteLength !== b.byteLength) return false;
+    return a.every((val, i) => val === b[i]);
```
tynes: I'm not sure what the metering model would be, but you could save gas by returning early here when a single element doesn't match instead of using `every`

js/utils/utils.js
```diff
@@ -0,0 +1,116 @@
...
+
+  safeSlice: (buf, first, last) => {
+    let start;
+    let end;
+
+    if (first === null || undefined) { start = 0; }
+    if (last === null || undefined) { end = buf.length - 1; }
+
+    if (typeof first === 'bigint') {
+      start = Number(first);
```
tynes: This is unsafe in some cases (for numbers larger than `Number.MAX_SAFE_INTEGER`) but unlikely in practice. I think it would be smarter to throw instead of allowing for unsafe behavior

```
> Number(900719925474099111n)
900719925474099100
```

js/utils/utils.js
```diff
@@ -0,0 +1,116 @@
...
+    return buf.slice(start, end);
+  },
+
+  /**
+   * @notice               JS version of abi.encodePacked when trying to concatenate 2 values
+   * @dev                  Use when you see abi.encodePacked
+   * @param {array}        a An array of Uint8Arrays
+   * @return {Uint8Array}  A Uint8Array that is a concatenation of all the arrays
+  */
+//  TODO: Can we make this so it takes in more than 2 arrays?
```
tynes: Is this what you mean?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

js/lib/BTCUtils.js
ErinHales: Oh, yes, I forgot that slice goes to that index, but doesn't include the value at that index.  I will fix.

ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+   */
+  validateHeaderChain: (headers) => {
+    // Check header chain length
+    if (headers.length % 80 !== 0) {
+      throw new Error('Header bytes not multiple of 80.');
+    }
+
+    // Initialize header start index
+    let digest;
+    let start = 0;
```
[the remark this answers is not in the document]
ErinHales: `start` can be moved inside, but `digest` should remain outside because `validateHeaderPrevHash` uses the hash of the previous block.  If digest is moved inside the for loop, it will reset each time instead of storing the previous hash.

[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+          throw new Error('Header bytes not a valid chain.');
+        }
+      }
+
+      // ith header target
+      let target = btcUtils.extractTarget(header);
+
+      // Require that the header has sufficient work
+      digest = btcUtils.hash256(header);
+      if (utils.bytesToUint(btcUtils.reverseEndianness(digest)) > target) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: remaining can be moved inside, but `len` needs to be outside so it can be referenced in the return statement.

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: This has already been fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: Ah, right, fixed

js/lib/BTCUtils.js
ErinHales: This has already been fixed.

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: Already fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: already fixed

js/lib/BTCUtils.js
ErinHales: already fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: already fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: fixed

js/lib/BTCUtils.js
ErinHales: already fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+  parseOutput: (output) => {
+    let value = btcUtils.extractValue(output);
+
+    if (utils.typedArraysAreEqual(utils.safeSlice(output, 9, 10), new Uint8Array([106]))) {
+      // OP_RETURN
+      outputType = OUTPUT_TYPES.OP_RETURN;
+      payload = btcUtils.extractOpReturnData(output);
+    } else {
+        let prefixHash = utils.safeSlice(output, 8, 10);
+        if (utils.typedArraysAreEqual(prefixHash, new Uint8Array([34, 0]))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+          // P2WSH
+          outputType = OUTPUT_TYPES.WSH;
+          payload = utils.safeSlice(output, 11, 43);
+        } else if (utils.typedArraysAreEqual(prefixHash, new Uint8Array([22, 0]))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+          // P2WPKH
+          outputType = OUTPUT_TYPES.WPKH;
+          payload = utils.safeSlice(output, 11, 31);
+        } else if (utils.typedArraysAreEqual(prefixHash, new Uint8Array([25, 118]))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+          // PKH
+          outputType = OUTPUT_TYPES.PKH;
+          payload = utils.safeSlice(output, 12, 32);
+        } else if (utils.typedArraysAreEqual(prefixHash, new Uint8Array([23, 169]))) {
```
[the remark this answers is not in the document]
ErinHales: fixed

js/lib/ValidateSPV.js
```diff
@@ -0,0 +1,231 @@
...
+          // SH
+          outputType = OUTPUT_TYPES.SH;
+          payload = utils.safeSlice(output, 11, 31);
+        } else {
+          outputType = OUTPUT_TYPES.NONSTANDARD;
+          payload = null;
```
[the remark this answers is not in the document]
ErinHales: fixed, also fixed test to check for empty array instead of null.

js/package.json
ErinHales: fixed

js/package.json
ErinHales: `{
  "engineStrict": true,
  "engines": {
    "node": "^10.0.0"
  },` Does that look correct?

js/utils/utils.js
ErinHales: Yes, something like that, thanks

ErinHales: ok, this is fixed

js/utils/utils.js
prestwich: ah you're probably right. something like this:

```
if (a.byteLength !== b.byteLength) return false;
if (a.BYTES_PER_ELEMENT !== b.BYTES_PER_ELEMENT) return false;
for (let i = 0; i < a.byteLength; i += 1) {
  if (a[i] != b[i]) return false;
}
return true;
```

js/utils/utils.js
```diff
@@ -0,0 +1,116 @@
...
+  concatUint8Arrays: (arrays) => {
```
prestwich: `concatUint8Arrays: (...arrays) => {`

then you can call it as `concatUint8Arrays(a,b,c,d,e)` instead of `concatUint8Arrays([a,b,c,d,e])`

ErinHales: This is showing old code, I've fixed this already :)

js/utils/utils.js
ErinHales: Updated

js/README.md
```diff
@@ -0,0 +1,131 @@
...
+It is extremely easy to write insecure code using these libraries. We do not
+recommend a specific security model. Any SPV verification involves complex
+security assumptions. Please seek external review for your design before
+building with these libraries.
+
+### Breaking changes from 1.x:
+
+- Merkle proof indexes are 0-indexed (like they should have been all along)
+- Merkle proofs for `ValidateSPV#prove` no longer require the leaf or root hash
+- Node 10.0.0 or greater is required
```
ErinHales: Added note about needing node 10 or greater here

js/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+While the prover is off-chain, and makes Ethereum transactions, the verifier is
+implemented as a solidity contract that validates proofs contained in those
+transactions. The verifier must set one parameter on-chain: the required total
+work, expressed as accumulated difficulty. The verifier sums difficulty across the header chain by measuring the work in its component headers.
+In addition, the verifier may set any number of other acceptance constraints
+on the proof. E.g. the contract may check that the `vout` contains an
+output paying at least 30,000 satoshi to a particular `scriptPubkey`.
+
+**Usage Example:**
```
ErinHales: Wrote new js usage example

js/test/utils.test.js
```diff
@@ -109,15 +169,86 @@ describe('utils', () => {
...
+      assert.isTrue(arraysAreEqual);
+
+      // slice with start index, but not end index
+      res = utils.safeSlice(arr, 2);
+      arraysAreEqual = utils.typedArraysAreEqual(res, new Uint8Array([3,4,5]));
+      assert.isTrue(arraysAreEqual);
+    });
+    it('error if passed invalid arguments', () => {
+      const arr = [1,2,3,4,5];
+      const OUT_OF_RANGE = BigInt(Number.MAX_SAFE_INTEGER + 1)
```
dominiquejane: `const OUT_OF_RANGE = BigInt(Number.MAX_SAFE_INTEGER) + BigInt(1);`

ErinHales: Is this necessary to change?  Both ways return the same value.

dominiquejane: The original way wasn't passing when I ran the tests.

prestwich: "max safe integer" is what it sounds like. Above that the behavior is unsafe, which is why it wouldn't work consistently

ErinHales: Oh, I understand now. Yes, you're right, I'll change this.

dominiquejane:
closes #26

PR #43, opened 2019-10-02
WIP: More usage examples

ErinHales:
- [x] Print the input/output type as a string instead of an integer
 - [x] Use Prove + CalculateTxID in a cli call to output (True) or (False+an error message)
        - This is basically finished.  Check that I'm understanding what is supposed to happen in `Prove`.  Also, I need to use a uint for Prove; I left a TODO in `main.go`.
 - [x] CLI for parse header and validate header chain


golang/btcspv/utils.go
```diff
@@ -14,10 +14,47 @@ func strip0xPrefix(s string) string {
 	return s
 }
 
+// Decodes a hex string into a byte array
```
prestwich: For Go, every exported function should have a comment that starts with the function name

`// DecodeIfHex decodes a hex string into a byte array`

golang/btcspv/utils.go
```diff
@@ -14,10 +14,47 @@ func strip0xPrefix(s string) string {
...
+		typeString = "PKH"
+	case 5:
+		typeString = "SH"
+	case 6:
+		typeString = "Nonstandard"
+	}
+	return typeString
+}
+
+func GetInputType(num uint) string {
```
prestwich: Needs comment

golang/btcspv/utils.go
```diff
@@ -14,10 +14,47 @@ func strip0xPrefix(s string) string {
...
+// Decodes a hex string into a byte array
 func DecodeIfHex(s string) []byte {
 	res, err := hex.DecodeString(strip0xPrefix(s))
 	if err != nil {
 		return []byte(s)
 	}
 	return res
 }
+
+func GetOutputType(num uint) string {
```
prestwich: needs comment

golang/examples/header.go
```diff
@@ -0,0 +1,54 @@
...
+		fmt.Println(err)
+	}
+	timeStr := time.Unix(unixIntValue, 0)
+
+	// Return data in a formatted string
+	dataStr := fmt.Sprintf("\nHeader #%d:\n  Digest: %s,\n  Version: %d,\n  Prev Hash: %s,\n  Merkle Root: %s,\n  Time Stamp: %s,\n  Target: %d,\n  Nonce: %d\n", num, digestStr, version, prevHashStr, merkleRootStr, timeStr, target, nonce)
+	return dataStr
+}
+
+func ParseHeader(header []byte) string {
```
prestwich: needs comment

golang/examples/header.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return "Error parsing header"
+	}
+
+	// Format data using prettifyHeaderData
+	headerData := prettifyHeaderData(0, digest, version, prevHash, merkleRoot, timestamp, target, nonce)
+	return headerData
+}
+
+func ValidateHeaderChain(headers []byte) string {
```
prestwich: needs comment

golang/examples/vout.go
```diff
@@ -7,9 +7,9 @@ import (
 	btcspv "github.com/summa-tx/bitcoin-spv/golang/btcspv"
 )
 
-func prettifyOutput(numOutput int, outpoint []byte, value uint, outputType btcspv.OutputType) string {
+func prettifyOutput(numOutput int, outpoint []byte, value uint, outputType string) string {
```
prestwich: don't change the `outputType` type here, keep it as `btcspv.OutputType`

golang/examples/vin.go
```diff
@@ -7,9 +7,9 @@ import (
 	btcspv "github.com/summa-tx/bitcoin-spv/golang/btcspv"
 )
 
-func prettifyInput(numInput int, outpoint []byte, index uint, inputType btcspv.InputType, sequence uint) string {
+func prettifyInput(numInput int, outpoint []byte, index uint, inputType string, sequence uint) string {
```
prestwich: don't change the `outputType` type here, keep it as `btcspv.OutputType`

golang/btcspv/utils.go
prestwich: change `num uint` to `outputType OutputType`

golang/btcspv/utils.go
```diff
@@ -14,10 +14,47 @@ func strip0xPrefix(s string) string {
...
+	var typeString string
+	switch num {
+	case 0:
```
prestwich: Actually, now that I think about it, we can use the enum names for the cases 

`case OutputNone`

golang/btcspv/utils.go
```diff
@@ -14,10 +14,47 @@ func strip0xPrefix(s string) string {
...
+	case 6:
+		typeString = "Nonstandard"
+	}
+	return typeString
+}
+
+func GetInputType(num uint) string {
+	var typeString string
+	switch num {
+	case 0:
```
prestwich: Actually, now that I think about it, we can use the enum names for the cases

`case InputNone`

prestwich:
@ErinHales `Prove.go` should be lowercase.

`git mv examples/Prove.go examples/prove.go`

golang/examples/main.go
```diff
@@ -15,6 +15,13 @@ func route(command string, arguments [][]byte) string {
 		result = ParseVin(arguments[0])
 	case "parseVout":
 		result = ParseVout(arguments[0])
+	case "parseHeader":
+		result = ParseHeader(arguments[0])
+	case "validateHeaderChain":
+		result = ValidateHeaderChain(arguments[0])
+	case "prove":
+		// TODO: fix arguments, 7th argument should be uint
```
prestwich: you'll need to change the bytes back to a string by something like `string(arguments[6])` and then use `strconv.ParseInt` [link](https://golang.org/pkg/strconv/#ParseInt)

golang/examples/Prove.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"errors"
+
+	btcspv "github.com/summa-tx/bitcoin-spv/golang/btcspv"
+)
+
+func Prove(version []byte, vin []byte, vout []byte, locktime []byte, merkleRoot []byte, intermediateNodes []byte, index uint) (string, error) {
```
prestwich: this doesn't need to return an error, for some reason, I thought that `btcsp.Prove`  bubbled up an error

golang/examples/header.go
```diff
@@ -0,0 +1,54 @@
...
+	// Get the total difficulty using ValidateHeaderChain
+	totalDifficulty, err := btcspv.ValidateHeaderChain(headers)
+	// Check for errors
+	if err != nil {
+		return "Error validating chain"
```
prestwich: `return fmt.Sprintf("%s", err)`

This will return whatever error message got generated :)

golang/examples/vout.go
ErinHales: Oh, yes, that should be done in the prettify functions

golang/btcspv/utils.go
ErinHales: Yes, that worked 🙂

golang/examples/header.go
ErinHales: Ok, had this problem in a couple of places, fixed all of them

golang/examples/main.go
ErinHales: Double check my work, but I think that's good

golang/examples/Prove.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+	"fmt"
+
+	btcspv "github.com/summa-tx/bitcoin-spv/golang/btcspv"
+)
+
+// Prove checks the validity of a merkle proof
+func Prove(version []byte, vin []byte, vout []byte, locktime []byte, merkleRoot []byte, intermediateNodes []byte, index uint) string {
```
prestwich: line breaks for lines longer than 80 please :)

golang/examples/header.go
```diff
@@ -0,0 +1,56 @@
...
+	"encoding/hex"
+	"fmt"
+	"strconv"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	btcspv "github.com/summa-tx/bitcoin-spv/golang/btcspv"
+)
+
+func prettifyHeaderData(num uint, digest []byte, version uint, prevHash []byte, merkleRoot []byte, timestamp uint, target sdk.Uint, nonce uint) string {
```
prestwich: line breaks for lines longer than 80 please :)

golang/examples/header.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Convert timestamp to readable time
+	unixIntValue, err := strconv.ParseInt(string(timestamp), 10, 64)
+	if err != nil {
+		return fmt.Sprintf("%s", err)
+	}
+	timeStr := time.Unix(unixIntValue, 0)
+
+	// Return data in a formatted string
+	dataStr := fmt.Sprintf("\nHeader #%d:\n  Digest: %s,\n  Version: %d,\n  Prev Hash: %s,\n  Merkle Root: %s,\n  Time Stamp: %s,\n  Target: %d,\n  Nonce: %d\n", num, digestStr, version, prevHashStr, merkleRootStr, timeStr, target, nonce)
```
prestwich: line breaks for lines longer than 80 please :)

golang/examples/main.go
```diff
@@ -15,6 +16,19 @@ func route(command string, arguments [][]byte) string {
...
+	case "parseHeader":
+		result = ParseHeader(arguments[0])
+	case "validateHeaderChain":
+		result = ValidateHeaderChain(arguments[0])
+	case "prove":
+		// convert argument to a uint
+		str := string(arguments[6])
+		uint64Arg, err := strconv.ParseUint(str, 10, 32)
+		if err != nil {
+			result = "Error converting arg to uint"
```
prestwich: need to `break` out of the switch case. Otherwise this error message will never be seen because we'll go straight to Prove afterwards

golang/examples/vout.go
```diff
@@ -38,7 +42,7 @@ func ParseVout(vout []byte) string {
 		voutData := prettifyOutput(numOutput, payload, value, outputType)
 
 		// Concat vout information onto `outputs`
-		outputs = outputs + "\n" + voutData
+		outputs = outputs + voutData
```
prestwich: names could be improved here. `outputs` could be something like `formattedOutputs` and `voutData` isn't actually about the `vout` at all. it's a single formatted output

same in `ParseVin`

ErinHales: Would renaming `voutData` to simply `data` suffice?

ErinHales: Or `outputData`?

PR #45, opened 2019-10-03
chore: move examples/ to cli/ and add build instructions to README

prestwich:


prestwich:
Feel free to add more after merge. Would also be good to get one that's not quite so chunky. :)

Issue #50, opened 2019-10-03
Standard JSON Proof format + utility functions


prestwich:
- Marshalling and Unmarshalling for each language (except solidity, obv.)
- Internal struct/object representation in each language
- Convenience passthrough to call e.g. `Prove` using the object/struct

Rules
1. Serialized bytes are hex
1. Hex should be `"0x"` prepended
1. No integers greater than `2**53` are required for proofs, so it's safe to use native `int`/`Number` types on 64-bit platforms
1. Structs and JS objects should have `Vec<u8>`, `[]byte`, `Uint8Array`, or language equivalent when deserialized

**Golang**
- [X]  structs  (#56 )
- [X]  json ser/deser (#56)
- [x] convenience functions (#60)

**JS**
- [x]  objects
- [x]  json ser/deser
- [x] convenience functions (#69)

**Solidity (covered by JS?)**
- [x]  objects
- [x]  json ser/deser
- [x] convenience functions (#69)
 
**Rust**
- [x] a library
- [x]  structs
- [x]  json ser/deser
- [x] convenience functions

**Python**
- [x]  TypedDicts
- [x]  json ser/deser
- [x] convenience functions (#70)

prestwich:
Note that the Python object here does not conform to rule 4

prestwich:
Done

PR #52, opened 2019-10-03
Golang json proofs

prestwich:
Needs tests.


prestwich:
in service of #50

golang/btcspv/types.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return err
+	}
+	copy(*h, buf)
+
+	return nil
+}
+
+// MarshallJSON marashalls 32 byte digests
+func (h HexBytes) MarshallJSON() ([]byte, error) {
```
ErinHales: I don't think this need to return an error.

golang/btcspv/types.go
```diff
@@ -0,0 +1,72 @@
...
+		return fmt.Errorf("Expected 32 bytes, got %d bytes", len(buf))
+	}
+
+	copy(h[:], buf)
+
+	return nil
+}
+
+// MarshallJSON marashalls 32 byte digests
+func (h Hash256Digest) MarshallJSON() ([]byte, error) {
```
ErinHales: Same here, don't think the error is needed, unless you plan to do something with it in the future.

golang/btcspv/types.go
prestwich: `MarshallJSON() ([]byte, error)` a standard interface invoked by a JSON framework. technically I don't think this _can_ error, but we still need to implement the interface so that the magic marshalling happens

golang/btcspv/types.go
prestwich: see above :+1:

PR #55, opened 2019-10-03
added address in prettifyOutput

### LINKED ISSUE #47: Add address associated with output payload
Add a line to `prettifyOutput` that prints the address associated with the output payload.

ErinHales:
Closes #47

### DIFF
--- a/golang/cli/vout.go
+++ b/golang/cli/vout.go
@@ -18,9 +18,24 @@ func prettifyOutput(
 	// Get the output type in readable format
 	outputTypeString := btcspv.GetOutputType(outputType)
 
+	// Get the address associated with the output
+	var address string
+	switch outputType {
+	case btcspv.WPKH:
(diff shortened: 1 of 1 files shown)


golang/cli/vout.go
```diff
@@ -18,9 +18,24 @@ func prettifyOutput(
 	// Get the output type in readable format
 	outputTypeString := btcspv.GetOutputType(outputType)
 
+	// Get the address associated with the output
+	var address string
+	switch outputType {
```
prestwich: Can we move this to a separate function?

ErinHales: Definitely, should I export it?

ErinHales: Ok, I just added it as an internal function.  Figured we could easily export it if we decide it would be useful to be accessible elsewhere.

PR #60, opened 2019-10-04
Add Validate Function

ErinHales:


golang/btcspv/types.go
```diff
@@ -28,11 +28,14 @@ type SPVProof struct {
 	Vin               HexBytes      `json:"vin"`
 	Vout              HexBytes      `json:"vout"`
 	Locktime          HexBytes      `json:"locktime"`
-	TxID              Hash256Digest `json:"tx_id"`
+	TxID              []byte        `json:"tx_id"`
```
prestwich: don't change types here, it'll break the serialization stuff

golang/btcspv/types.go
```diff
@@ -28,11 +28,14 @@ type SPVProof struct {
...
 	TxIDLE            Hash256Digest `json:"tx_id_le"`
 	Index             uint32        `json:"index"`
-	ConfirmingHeader  BitcoinHeader `json:"confirming_header"`
+	ConfirmingHeader  []byte        `json:"confirming_header"`
```
prestwich: same here

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,32 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
...
+
+func (s SPVProof) Validate() (bool, error) {
+	// Calculate the Tx ID and compare it to the one in SPVProof
+	txid := CalculateTxID(s.Version, s.Vin, s.Vout, s.Locktime)
+	if bytes.Compare(txid, s.TxID) != 0 {
+		return false, errors.New("Version, Vin, Vout and Locktime did not yield correct TxID")
+	}
+
+	// Get the merkle root, needed for Prove function
+	_, _, _, merkleRoot, _, _, _, err := ParseHeader(s.ConfirmingHeader)
```
prestwich: You should be able to use `s.ConfirmingHeader.MerkleRootLE` instead of parsing bytes

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,32 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	// Check the merkle proof
+	merkleProof := Prove(s.TxID, merkleRoot, s.IntermediateNodes, uint(s.Index))
```
prestwich: better variable name: `proofValid`

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,32 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
...
+	if !merkleProof {
+		return false, errors.New("Not a valid Merkle Proof")
+	}
+
+	// Validate the header chain, looks like the SPVProof doesn't have a header chain, only ConfirmingHeader
+	_, validationErr := ValidateHeaderChain(s.ConfirmingHeader)
```
prestwich: use `s.ConfirmingHeader.Raw`

prestwich: we don't neeed to check a whole chain in this particular step

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,32 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
 	}
 	return totalDifficulty, nil
 }
+
+func (s SPVProof) Validate() (bool, error) {
+	// Calculate the Tx ID and compare it to the one in SPVProof
+	txid := CalculateTxID(s.Version, s.Vin, s.Vout, s.Locktime)
+	if bytes.Compare(txid, s.TxID) != 0 {
```
prestwich: Need to use the LE txid here.
Also try casting the txid to a bytearray to solve your type issue:

`if !bytes.Equal(txid, []byte(s.TxIDLE))`

golang/btcspv/validate_spv.go
prestwich: need to use `s.TxIDLE` and `s.ConfirmingHeader.MerkleRootLE` here

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,32 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
 	}
 	return totalDifficulty, nil
 }
+
+func (s SPVProof) Validate() (bool, error) {
+	// Calculate the Tx ID and compare it to the one in SPVProof
+	txid := CalculateTxID(s.Version, s.Vin, s.Vout, s.Locktime)
```
prestwich: Add a line here: 
```
validHeader, err := s.ConfirmingHeader.Validate()
if err != nil {
  return false, err
}
```


Then add a Validate fuction to `BitcoinHeader` that checks:
1. `bytes.Equal(Hash256(b.Raw), b.HashLE)`
2. `b.Hash` is `b.HashLE` backwards
3. same for `b.MerkleRoot`
4. `bytes.Equal(b.MerkleRootLE, ExtractMerkleRootLE(b.Raw))`
5. `bytes.Equal(b.PrevHash, ExtractPrevBlockHashLE(b.Raw))`

golang/btcspv/validate_spv.go
prestwich: you probably actually need to do 
`if !bytes.Equal(txid, s.TxIDLE[:])`

golang/btcspv/types.go
ErinHales: Fixed

golang/btcspv/validate_spv.go
ErinHales: Ah, I see how that works now

prestwich:
A bunch of things are resolved, is there an unpushed commit?

golang/btcspv/validate_spv.go
prestwich: need to add the call here, so that we validate the header when we validate the proof object

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,80 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
 	}
 	return totalDifficulty, nil
 }
+
+func (b BitcoinHeader) Validate() (bool, error) {
```
prestwich: Need a comment here `// Validate .......`

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,80 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
...
+	// Check that PrevHash is the correct PrevHash for the header
+	validPrevHash := bytes.Equal(prevHash, ExtractPrevBlockHashLE(header))
+	if !validPrevHash {
+		return false, errors.New("Prev hash is not the correct previous hash of the header")
+	}
+
+	return true, nil
+}
+
+func (s SPVProof) Validate() (bool, error) {
```
prestwich: Need a comment here `// Validate .......`

ErinHales:
@prestwich Tests are finished, but the `CalculateTxID` in `Validate` for the SPV Proof is failing.  I think the data in `testProofs.json` may be wrong, could you check that out? ```"{\"version\": \"0x01000000\",\"vin\": \"0x0101748906a5c7064550a594c4683ffc6d1ee25292b638c4328bb66403cfceb58a000000006a4730440220364301a77ee7ae34fa71768941a2aad5bd1fa8d3e30d4ce6424d8752e83f2c1b02203c9f8aafced701f59ffb7c151ff2523f3ed1586d29b674efb489e803e9bf93050121029b3008c0fa147fd9db5146e42b27eb0a77389497713d3aad083313d1b1b05ec0ffffffff\", \"vout\": \"0x0116312f00000000001976a91400cc8d95d6835252e0d95eb03b11691a21a7bac588ac\",\"locktime\": \"0x00000000\",\"tx_id\": \"0x74d6d6dc1fc9b0f393abde12e76adeeb3d674b38b7fbea4d9fc28b3bb0f67651\"```

prestwich:
Now just need to add tests for the error cases :)

golang/btcspv/types_test.go
```diff
@@ -38,6 +52,45 @@ func TestTypes(t *testing.T) {
...
 	typesSuite.Fixtures = fixtures
 
+	spvProof := new(SPVProof)
+	err = json.Unmarshal([]byte(typesSuite.Fixtures.Valid[0]), &spvProof)
+	logIfErr(err)
+	typesSuite.Proof = *spvProof
+
+	// TODO: use length of typesSuite.BadHeaders instead of using 5
+	// for i := range typesSuite.BadHeaders {
+	for i := 0; i < 5; i++ {
```
ErinHales: What is the best way to write this line using the length of `BadHeaders`? `for i := range typesSuite.BadHeaders` gave an `Index out of range` error.  And `for i := 0; i < typesSuite.BadHeaders.len; i++` gave a `len is not defined` error.

golang/btcspv/validate_spv.go
```diff
@@ -184,3 +184,90 @@ func ValidateHeaderChain(headers []byte) (sdk.Uint, error) {
...
+	}
+
+	// Check that the proof is valid
+	validProof := Prove(txIDLE, merkleRootLE, intermediateNodes, index)
+	if !validProof {
+		return false, errors.New("Not a valid Merkle Proof")
+	}
+
+	// Validate the header chain, looks like the SPVProof doesn't have a header chain, only ConfirmingHeader
+	_, validationErr := ValidateHeaderChain(rawHeader)
```
ErinHales: Not sure how to test this function.  If I give it a header that is not valid, it gets caught by line 254 where we validate the header.

prestwich: Can just drop this block, as the current SPV proof format does not allow for chains

PR #67, opened 2019-10-12
Restructure data

ErinHales:


golang/btcspv/types_test.go
```diff
@@ -55,38 +57,26 @@ func TestTypes(t *testing.T) {
...
-		typesSuite.BadHeaders = appended
-	}
 
-	for i := 0; i < len(typesSuite.Fixtures.BadHeaderErrors); i++ {
-		headerErr := new(Err)
-		err = json.Unmarshal([]byte(typesSuite.Fixtures.BadHeaderErrors[i]), &headerErr)
-		logIfErr(err)
-		appended := append(typesSuite.BadHeaderErrors, *headerErr)
-		typesSuite.BadHeaderErrors = appended
+		invalidHeader.Error = strings.Replace(invalidHeader.Error, "\u00a0", " ", -1)
```
ErinHales: In the invalid header cases, the error strings come back with regex, i.e. `"MerkleRootLE\u00a0is\u00a0not\u00a0the\u00a0LE\u00a0version\u00a0of\u00a0MerkleRoot"`.  This messes with tests so I had to filter them out.  I don't understand why we have this problem here, but not with the invalid proof cases.

ErinHales:
I couldn't figure out how to simplify the Unmarshalling logic like you wanted.  The data wasn't making it through right, almost like it just made an empty struct rather than using the data from the json. ```{btcspv.InvalidHeadersCases{Error:"", Header:btcspv.BitcoinHeader{Raw:btcspv.RawHeader{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, Hash:btcspv.Hash256Digest{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0...```

PR #82, opened 2019-11-12
Add Types Tests

ErinHales:


golang/btcspv/types_test.go
```diff
@@ -171,3 +171,60 @@ func (suite *TypesSuite) TestValidateSPVProof() {
...
+
+	header, err := NewRawHeader(input)
+	suite.Nil(err)
+	suite.Equal(header, output)
+}
+
+func (suite *TypesSuite) TestHeaderFromRaw() {
+	validHeader := suite.ValidProof.ConfirmingHeader
+	// TODO: Why is the prevhash the reverse endianness of what it should be?
+	validHeader.PrevHash = Hash256Digest{0xc2, 0x38, 0xb6, 0x1, 0x30, 0x8b, 0x72, 0x97, 0x34, 0x6a, 0xb2, 0xed, 0x59, 0x94, 0x2d, 0x7d, 0x7e, 0xce, 0xa8, 0xd2, 0x3a, 0x10, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}
```
ErinHales: For some reason, `PrevHash` is the reverse endianness of what it should be.  Not sure what is causing this or how to reverse endianness of `Hash256Digest`.  For now, I just explicitly assigned it to the correct value.

prestwich: it may be a bug in my implementation. what endianness is it supposed to be, and what endianness is it?

Come to think of it, this is the only hash field that is not included both ways and explicitly labeled. Might be a good idea to fix that :thinking:

ErinHales: I believe it's supposed to be LE.

ErinHales: This would be little endian, right? `0x00000000000000000001103ad2a8ce7e7d2d9459edb26a3497728b3001b638c2`

prestwich: > `0x00000000000000000001103ad2a8ce7e7d2d9459edb26a3497728b3001b638c2`

this is BE

ErinHales: Ok, I think I know  the problem then.  How would I reverse endianness of type `Hash256Digest`.  `ReverseEndianness` takes a byte array and I can't figure out how to convert `Hash256Digest` to a byte array.

ErinHales: In JSON, the prev hash is BE, so I need to change that to LE before comparing.

prestwich: check what I did in the `Verify` functions

Also, write a convenience function to do this so I stop doing it inline.

```
func ReverseHash256Endianness(h Hash256Digest) Hash256Digest {}
```

ErinHales: Never mind.  Figured it out

ErinHales: Sorry, I just saw your last comment.  Yeah, I can make a function to do it.

prestwich:
missing coverage on line 120 of `types.go`. Pass in a string containing non-hex characters

golang/btcspv/types_test.go
```diff
@@ -171,3 +171,78 @@ func (suite *TypesSuite) TestValidateSPVProof() {
...
+	validHeader.PrevHash = reversed
+	var height uint32 = 592920
+
+	rawHeader := HeaderFromRaw(validHeader.Raw, height)
+	suite.Equal(rawHeader, validHeader)
+}
+
+func (suite *TypesSuite) TestHeaderFromHex() {
+	hex := "0x0000c020c238b601308b7297346ab2ed59942d7d7ecea8d23a1001000000000000000000b61ac92842abc82aa93644b190fc18ad46c6738337e78bc0c69ab21c5d5ee2ddd6376d5d3e211a17d8706a84"
+	var height uint32 = 592920
```
prestwich: can we get this from the test suite instead of hardcoding?

traverse the valid header array and make sure each one can be generated from its own `raw` and `height` fields?

prestwich: Same goes for all hard-coded constants

golang/btcspv/types_test.go
ErinHales: Ok, made the function, a test for the function, and replaced all instances that try to do this inline

golang/btcspv/types_test.go
ErinHales: Right now we only have one valid header, but I can make it so if we add more, it will test all of them.

ErinHales: Fixed this, now getting these values from the test suite

PR #89, opened 2019-11-12
PrevhashLE - JS

ErinHales:
- Clarified prevhash type in several places


testVectors.json
```diff
@@ -651,7 +651,7 @@
       "output": {
         "digest": "0x00000000000000000024cc6777e93673f53853240d34f1bb7fb1d63983e470fe",
         "version": 536870912,
-        "prevHash": "0x73bd2184edd9c4fc76642ea6754ee40136970efc10c419000000000000000000",
```
prestwich: does this change break tests in other languages?

ErinHales: Good point.  Yeah, I'll change it back

PR #92, opened 2019-11-16
cli: port from hackathon

tynes:
Cleans up and ports the cli from the hackathon


tynes:
I realized that the `headers` command doesn't return json. Its kind of weird that `proof` returns json while `headers` does not

tynes:
Do the docs make sense?

bin/BcoinClient.js
```diff
@@ -0,0 +1,281 @@
...
+
+      if (enc === 'json') {
+        headers.push(json);
+      } else {
+        const hex = await this.execute('getblockheader', [json.hash, false]);
+        headers.push(hex);
+      }
+    }
+
+    if (enc === 'hex')
```
tynes: This should probably return `{headers: headers.join('')}` so that return types are consistently json

bin/BcoinClient.js
```diff
@@ -0,0 +1,281 @@
...
+
+function reverse(str) {
+  return Buffer.from(str, 'hex').reverse().toString('hex');
+}
+
+/**
+ * Parse a hex transaction into an object
+ */
+
+function parseTxHex(hex) {
```
tynes: Do you want to move this into `bitcoin-spv/js/utils` or something like that? If so, it could use some tests. Pretty sure it works tho

barbaraliau: I think that would be useful.

bin/BcoinClient.js
barbaraliau: for line 168 and 170?

tynes: yup

bin/BcoinClient.js
tynes: @prestwich was mentioning a clean split between "on chain" and "off chain" code. This would go in the off chain section, although I could see it going on chain (in a rewritten form not using string concatenation, just raw Uint8Arrays) if a different contract was to call the relay contracts

barbaraliau:
> Do the docs make sense?

Yes. Always good to have more examples though.

Since BcoinClient can be used without the cli, should it be moved from `bin/BcoinClient.js` to `/clients/BcoinClient.js` and require into `bin/index.js` from there?

bin/BcoinClient.js
barbaraliau: Yeah. Same return type is 👍

tynes:
> > Do the docs make sense?
> 
> Yes. Always good to have more examples though.
> 
> Since BcoinClient can be used without the cli, should it be moved from `bin/BcoinClient.js` to `/clients/BcoinClient.js` and require into `bin/index.js` from there?

Yeah, that makes sense. Note that I bundled up the few small deps `BcoinClient` needs and vendored them in, so they would have to be moved as well. You can submit a PR against my branch or just merge then fix in a new PR

barbaraliau:
> > > Do the docs make sense?
> > 
> > 
> > Yes. Always good to have more examples though.
> > Since BcoinClient can be used without the cli, should it be moved from `bin/BcoinClient.js` to `/clients/BcoinClient.js` and require into `bin/index.js` from there?
> 
> Yeah, that makes sense. Note that I bundled up the few small deps `BcoinClient` needs and vendored them in, so they would have to be moved as well. You can submit a PR against my branch or just merge then fix in a new PR

@tynes https://github.com/tynes/bitcoin-spv/pull/1

tynes:
@barbaraliau 

- renamed `bin/index.js` to `bin/cli` and made it executable
- updated `README` with more examples, 1 for each of the two commands

If the updated `README` looks good, I think its ready to merge

barbaraliau:
@tynes I like the examples. 👍 LGTM. Merging.

PR #95, opened 2019-11-21
Remove hex dependency in btcutils

### LINKED ISSUE #84: Remove `hex` dependency from `bitcoin_spv.go`
I don't like that we rely on string (de)serialization in bitcoin spv. we should figure out alternative elegant ways of doing this

ErinHales:
Closes #84


prestwich:
L363 of `bitcoin_spv.go` depends on `hex`

ErinHales:
Riiight, ok, that is fixed now.

PR #99, opened 2019-12-02
Update client path, docs, add additional methods

barbaraliau:


js/clients/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+```js
+const { BcoinClient } = require('@summa-tx/bitcoin-spv-js-clients');
+```
+
+ES6 module:
+
+```js
+import { BconClient } from '@summa-tx/bitcoin-spv-js-clients';
```
prestwich: bcon

js/clients/README.md
```diff
@@ -0,0 +1,184 @@
...
+// [
+//   '8e2d404a039a7a3e1768b161aa23546aab0444b73905bdd3d68b3d6f1769e8c0...',
+//   4
+// ]
+```
+
+#### client.getHeaderChainByCount(height, count, enc)
+* `@param {Number} height` - starting block height
+* `@param {Number} count` - number of headers
+* `@param {String} enc` - json or hex
```
prestwich: also supports `btcspv now`

js/clients/lib/BcoinClient.js
```diff
@@ -143,6 +167,7 @@ class BcoinClient extends NodeClient {
    * @param {Number} height - starting block height
    * @param {Number} count - number of headers
    * @param {String} enc - json or hex
```
prestwich: also `btcspv` now

js/clients/README.md
barbaraliau: akon

barbaraliau:
Ok I made changes. And fixed the history.

js/clients/README.md
```diff
@@ -0,0 +1,184 @@
...
+const client = new BcoinClient(options);
+const block = 12345;
+const header = client.getHeader(block);
+
+// {
+//   raw: hex,
+//   hash: json.hash,
+//   hash_le: reverse(json.hash),
+//   height: typeof block === 'number' ? block : json.height,
+//   prevhash: json.prevBlock,
```
prestwich: this should include `prevhash_le` now as well

js/clients/lib/BcoinClient.js
```diff
@@ -1,16 +1,23 @@
...
+* @copyright (c) [Summa]{@link https://summa.one} 2019
+* @module clients
+*
+*/
 
 'use strict';
 
-const {NodeClient} = require('./vendor/bclient');
+const { NodeClient } = require('./vendor/bclient');
 const assert = require('./vendor/bsert');
```
prestwich: these two vendors are broken

js/clients/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "version": "0.1.0",
+  "description": "Clients for fetching bitcoin SPV proofs in Javascript",
+  "main": "lib/index.js",
+  "engines": {
+    "node": ">10.0.0"
+  },
+  "engineStrict": true,
+  "scripts": {
+    "lint": "eslint ./test ./src",
+    "lint:fix": "eslint --fix ./test ./src",
```
prestwich: these don't lint anything in this package

js/clients/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "test": "mocha test -r esm",
+    "test:coverage": "jest --coverage",
```
prestwich: no tests in this package

js/clients/lib/BcoinClient.js
barbaraliau: Fixed

js/clients/package.json
barbaraliau: Fixed

js/clients/package.json
barbaraliau: Fixed

barbaraliau:
Realized that there are several variables that are not declared. Need to be ventured in.

barbaraliau:
@prestwich variables come from using `bcoin` in `getHeaderChain(height, nwork)`. This is the method that was originally used when the `BcoinClient` was in `bitcoin-verifier-ses`.

Can this be replaced with `getHeaderChainByCount()`? This will change the use of predicates that use `nwork`.

prestwich:
> Can this be replaced with getHeaderChainByCount()? This will change the use of predicates that use nwork.

unfortunately no. they have very different purposes

barbaraliau:
Ok. This was the one method I didn't rewrite with no dependencies. I will
vendor or rewrite then.

tynes:
If you lint the vendored deps, then its going to be a pain to diff them against new versions

js/clients/README.md
```diff
@@ -0,0 +1,185 @@
...
+const header = client.getMerkleProof(txid, height);
+
+// merkle proof
+// [
+//   '8e2d404a039a7a3e1768b161aa23546aab0444b73905bdd3d68b3d6f1769e8c0...',
+//   4
+// ]
+```
+
+#### client.getHeaderChainByCount(height, count, enc)
```
tynes: The concept of passing an encoding type is a little bit vestigial at this point, I originally had it because sometimes we wanted to pretty print and sometimes we wanted the raw array buffers. I think it may be a good idea to eliminate the `enc` variable to simplify and always return json (or uint8arrays if preferred) and then the ser package can be used to convert between the two.

js/clients/lib/BcoinClient.js
```diff
@@ -19,12 +25,11 @@ const {utils, BTCUtils} = require('../js/dist');
  */
 class BcoinClient extends NodeClient {
   /**
-   * bcoinclient constructor.
+   * bcoinclient constructor
+   * NB: none since constructor doesn't do anything
+   * (see: eslint-no-useless-constructor)
    * @param {object} options
    */
-  constructor(options) {
```
tynes: It will automatically call the `NodeClient` constructor and pass along the arguments? Didn't know that

js/clients/README.md
barbaraliau: Yeah that makes sense. I need to finish up this refactor and push all the changes I have.

js/clients/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### client.getHeader(block)
+* `@param {Hash|Number} block` - block height or hash
+* `@returns {Object} header` - header object
+
+Gets header by height or hash.
+```js
+const client = new BcoinClient(options);
+const block = 12345;
+const header = client.getHeader(block);
```
tynes: Missing `await`

js/clients/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### client.getProof(txid)
+   * `@param {String} txid` - txid in big endian
+   * `@returns {Object}` - proof object
+
+Get a proof for a transaction.
+
+```js
+const client = new BcoinClient(options);
+const txid =
```
tynes: Missing `txid` value

js/clients/README.md
```diff
@@ -0,0 +1,185 @@
...
+//       "version": 536870912,
+//       "prevBlock": "0c4ea5e675941eca1909275f21903cef755069a03c57c10f4d4dadcdd7146daf",
+//       "merkleRoot": "9a249c682aba07943c8a1f9bd774a15d71372fcd7f3f9ee99e8c7aa022ae6aa0",
+//       "time": 1571661863,
+//       "bits": 545259519,
+//       "nonce": 8,
+//       "height": 100,
+//       "chainwork": "00000000000000000000000000000000000000000000000000000000000000ca"
+//     },
+//         {
```
tynes: Spacing is off here

js/clients/lib/BcoinClient.js
barbaraliau: If there isn't one specified, it provides a default one. For derived classes:

```
constructor(...args) {
  super(...args);
}
```

I don't like unnecessary code, so it's kind of nice. If we decide we want to put them in at a later time then we can always disable the rule and add them back in.

js/clients/README.md
barbaraliau: Fixing.

barbaraliau: And all the other ones.

tynes:
Sorry about removing some of the methods that are needed for the ses code when I was porting, was only thinking about the CLI

barbaraliau:
> If you lint the vendored deps, then its going to be a pain to diff them against different versions. Granted the vendored deps should be pretty stable, but you never know.

I added an `.eslintignore` so those files aren't linted. For the above reason, but also that would be a pain to lint lol. It was something like 323 linting errors.

barbaraliau:
> Sorry about removing some of the methods that are needed for the ses code when I was porting, was only thinking about the CLI

@tynes No worries. I didn't realize it was needed either until I was rewriting the `README.md` for the ses code, and then realized that I needed to update all the `proof` objects to make sure that the client was returning the right type, since almost all of them need to be `Uint8Array`s

js/clients/README.md
barbaraliau: Going to return `Uint8Array`s for all of these since the primary use case is for the verifier code, and we can use the `ser` package to go back and forth as needed.

tynes:
> @tynes No worries. I didn't realize it was needed either until I was rewriting the `README.md` for the ses code, and then realized that I needed to update all the `proof` objects to make sure that the client was returning the right type, since almost all of them need to be `Uint8Array`s

So the plan is to delete the client in the ses code and then use the one here?

tynes:
Also see this commit https://github.com/summa-tx/bitcoin-spv/commit/559471b6c64a98ff42acdd1901b4266a2c77073f, I removed methods like `getHeadersByWork` and `getHeadersByCurrency`.

`getHeadersByWork` was meant to return an amount of headers that satisfies `nwork` (BN representation of work), summing each header's work together until surpasses `nwork`

Since thinking in `nwork` is not intuitive, `getHeadersByCurrency` was supposed to take a dollar amount and then use an external API to fetch the price of BTC and then try to calculate a dollar based amount of headers, so that you can say "this headers chain is 1 million dollars worth of security"

barbaraliau:
> > @tynes No worries. I didn't realize it was needed either until I was rewriting the `README.md` for the ses code, and then realized that I needed to update all the `proof` objects to make sure that the client was returning the right type, since almost all of them need to be `Uint8Array`s
> 
> So the plan is to delete the client in the ses code and then use the one here?

TL;DR, yes. I was refactoring the client in the ses code, but then this code was pushed up and since we were, trying to remove as many dependencies as possible from the ses code, it made sense to finish this up and use it, since it was almost identical, barring one or two methods.

Side note: I wonder if we could pass in the preferred encoding when instantiating the client, as a nice way of saying I want all my methods to return this type, without having to specify it in each method, and use the `ser` package internally to do that.

js/clients/README.md
barbaraliau: Jk. I think I'll save this for a different time. For now use the `ser` package on the other side.

tynes:
> Side note: I wonder if we could pass in the preferred encoding when instantiating the client, as a nice way of saying I want all my methods to return this type, without having to specify it in each method, and use the ser package internally to do that.

I like this idea, different applications would probably want the results in a particular format. The CLI wants all json and the ses code wants all uint8arrays

prestwich:
Where are we on this?

PR #118, opened 2020-02-14
bug: fix potential overflow in determineOutputLength

prestwich:


prestwich:
Hold off until we test new vectors in all languages

PR #119, opened 2020-02-14
bug: fix underflow in extractHash

prestwich:
Hold off until we test new vectors in all languages


solidity/contracts/BTCUtils.sol
```diff
@@ -368,7 +368,11 @@ library BTCUtils {
     /// @return          The hash committed to by the pk_script, or null for errors
     function extractHash(bytes memory _output) internal pure returns (bytes memory) {
         if (uint8(_output.slice(9, 1)[0]) == 0) {
-            uint256 _len = uint8(extractOutputScriptLen(_output)[0]) - 2;
+            uint256 _len = uint8(extractOutputScriptLen(_output)[0]);
+            if (_len < 2) {
+              return hex"";
+            }
+            _len -= 2;
```
ErinHales: This is a tad more concise.
```
uint256 _len = uint8(extractOutputScriptLen(_output)[0]) - 2;
if (_len < 0) {
    return hex"";
}
```

prestwich: It's an unsigned integer, so this would undeflow. The actual check would be `_len > 0xfd`. and that's less legible than what I have

ErinHales: Oh, got it

PR #121, opened 2020-02-18
Merkle Index docs

### LINKED ISSUE #108: `verifyHash256Merkle` allows existence proofs for the same leaf in multiple locations in the tree
#### Description

`BTCUtils.verifyHash256Merkle` is used by `ValidateSPV.prove` to validate a transaction's existence in a Bitcoin block. The function accepts as input a `_proof` and an `_index`. The `_proof` consists of, in order: the transaction hash, a list of intermediate nodes, and the merkle root.

The proof is performed iteratively, and uses the `_index` to determine whether the next proof element represents a "left branch" or a "right branch:"

https://github.com/summa-tx/bitcoin-spv/blob/f5e4da091a1c97e6432c2d70eba434edb189f919/solidity/contracts/BTCUtils.sol#L574-L586

If `_idx` is even, the computed hash is placed before the next proof element. If `_idx` is odd, the computed hash is placed after the next proof element. After each iteration, `_idx` is decremented by `_idx /= 2`.

Because `verifyHash256Merkle` makes no requirements on the size of `_proof` relative to `_index`, it is possible to pass in invalid values for `_index` that prove a transaction's existence in multiple locations in the tree.

#### Examples

By modifying existing tests, we showed that any transaction can be proven to exist at at least one alternate index. This alternate index is calculated as `(2 ** treeHeight) + prevIndex` - though other alternate indices are possible. The modified test is below:

```
it('verifies a bitcoin merkle root', async () => {
  for (let i = 0; i < verifyHash256Merkle.length; i += 1) {
    const res = await instance.verifyHash256Merkle(
      verifyHash256Merkle[i].input.proof,
      verifyHash256Merkle[i].input.index
    ); // 0-indexed
    assert.strictEqual(res, verifyHash256Merkle[i].output);

    // Now, attempt to use the same proof to verify the same leaf at
    // a different index in the tree:
    let pLen = verifyHash256Merkle[i].input.proof.length;
    let height = ((pLen - 2) / 64) - 2;

    // Only attempt to verify roots that are meant to be verified
    if (verifyHash256Merkle[i].output && height >= 1) {
      let altIdx = (2 ** height) + verifyHash256Merkle[i].input.index;

      const resNext = await instance.verifyHash256Merkle(
        verifyHash256Merkle[i].input.proof,
        altIdx
      );

      assert.strictEqual(resNext, verifyHash256Merkle[i].output);

      console.log('Verified transaction twice!');
    }
  }
});
```

#### Recommendation

Use the length of `_proof` to determine the maximum allowed `_index`. `_index` should satisfy the following criterion: `_index < 2 ** (_proof.length.div(32) - 2)`.

Note that subtraction by 2 accounts for the transaction hash and merkle root, which are assumed to be encoded in the proof along with the intermediate nodes.

ErinHales:
Closes #108


c/csrc/btcspv.h
```diff
@@ -186,7 +186,7 @@ byte_view_t btcspv_extract_input_tx_id_le(const_view_t *tx_in);
 /// @param input     The input
 /// @warning         overwrites `hash`
 /// @warning         caller must ensure `hash` is allocated and can hold 32 bytes
-void btcspv_extract_extract_input_tx_id_be(uint256 hash, const_view_t *tx_in);
```
prestwich: needs to be changed in btcspv.c as well

PR #127, opened 2020-02-18
Remove be and parse

prestwich:
Remove the least useful functions from the on-chain code. 

Includes #124

PR #128, opened 2020-02-19
Support multi-byte VarInts throughout

prestwich:
Includes #127 and #124.


prestwich:
TODO: rewrite extract input at index in all languages

golang/btcspv/types.go
```diff
@@ -102,16 +77,19 @@ func HeaderFromRaw(raw RawHeader, height uint32) BitcoinHeader {
 	digestLE := Hash256(raw[:])
 	digestBE := ReverseHash256Endianness(digestLE)
 	prevhashLE := ExtractPrevBlockHashLE(raw)
-	prevhash := ReverseHash256Endianness(prevhashLE)
+	prevhashBE := ReverseHash256Endianness(prevhashLE)
+  merkleRootLE := ExtractMerkleRootLE(raw)
```
ErinHales: Match indentation

solidity/contracts/BTCUtils.sol
```diff
@@ -137,20 +157,29 @@ library BTCUtils {
...
     /// @return          The input as a byte array
-    function extractInputAtIndex(bytes memory _vin, uint8 _index) internal pure returns (bytes memory) {
-        uint256 _len;
+    function extractInputAtIndex(bytes memory _vin, uint256 _index) internal pure returns (bytes memory) {
+        uint256 _varIntDataLen;
+        uint256 _nIns;
+
+        (_varIntDataLen, _nIns) = parseVarInt(_vin);
+        require(_varIntDataLen != ERR_BAD_ARG, "Read overrun during VarInt parsing");
+        require(_index <= _nIns, "Vin read overrun");
```
wadeAlexC: Note that these requires can be bypassed if `determineVarIntDataLength(_vin)` is zero and `_index` is zero.

solidity/contracts/BTCUtils.sol
```diff
@@ -137,20 +157,29 @@ library BTCUtils {
...
+        uint256 _len = 0;
+        uint256 _offset = 1 + _varIntDataLen;
 
-        for (uint8 _i = 0; _i < _index; _i ++) {
+        for (uint256 _i = 0; _i < _index; _i ++) {
             _remaining = _vin.slice(_offset, _vin.length - _offset);
             _len = determineInputLength(_remaining);
+            require(_len != ERR_BAD_ARG, "Bad VarInt in scriptSig");
             _offset = _offset + _len;
         }
```
wadeAlexC: In this case, this for loop is skipped

solidity/contracts/BTCUtils.sol
```diff
@@ -137,20 +157,29 @@ library BTCUtils {
...
 
         _remaining = _vin.slice(_offset, _vin.length - _offset);
         _len = determineInputLength(_remaining);
+        require(_len != ERR_BAD_ARG, "Bad VarInt in scriptSig");
         return _vin.slice(_offset, _len);
     }
```
wadeAlexC: And `_vin.slice(1, _vin.length - 1)` returns what may be a totally valid input

prestwich: Yes, this is the intended functionality. If `index` is 0, it should skip the loop.

solidity/contracts/BTCUtils.sol
prestwich: slight bug here. 166 should be `<` instead of `<=`. same for L343 in `extractOutputAtIndex`

PR #135, opened 2020-03-25
Organized test files

ErinHales:
The `btcspv` folder was getting messy, so I organized the test files into their own folder.  I made this into its own PR so I could get eyes on it early on and make sure this is a structure that we liked.
 - Moves test files into `btcspv/tests` folder and creates new `tests` package.
 - Exports more functions from `btcspv` including `Hash256MerkleStep`, `Strip0xPrefix` and `ZeroBytesError` so they can be tested from the `tests` package.


golang/btcspv/bitcoin_spv.go
```diff
@@ -479,7 +479,7 @@ func ExtractDifficulty(header RawHeader) sdk.Uint {
 	return CalculateDifficulty(ExtractTarget(header))
 }
 
-func hash256MerkleStep(a []byte, b []byte) Hash256Digest {
```
prestwich: needs an `// Hash256MerkleStep ......` comment if it's going to be exported

ErinHales:
I see what you're saying about the coverage tool being broken now.  I'll see what I can do to fix it.

ErinHales:
Hm, after some research, I think I am going to have to abandon this PR.  Reasons being:
 - Coverage simply doesn't work from another folder/package; Go just wasn't designed that way.
 - If I move just the `_test.go` files back and leave the `_test_type.go` files inside the test folder, I run into an import cycle which Go doesn't like either (test_utils imports btcspv and btcspv imports test_utils).

With these two issues, it leaves us back where we started with all the files in the same `btcspv` folder.  One last idea to clean up the files would be to make a singular `test_types.go` file.  But that also makes for one rather large file...

prestwich:
I guess this could be solved by moving `types.go` into its own package?

so 
```
golang - types/ - types.go
       |        |
       |        | - types_test.go
       |        |
       |        - *_test_types.go
       |
       | - bitcoin_spv.go
       | - bitcoin_spv_test.go
       - - etc.
```

ErinHales:
Hm, that would almost work, except `types.go` uses several functions from btcspv.  So we would just end up having an import cycle there.  Am I missing something?  I feel like this shouldn't be that big of a problem...

prestwich:
ya you're right. all the structs use the bitcoin_spv stuff :(

prestwich:
the real solution would be to break up types to linearize the inheritance :thinking: 

make `primitives/` which has low-level things like `RawHeader` and `Hash256Digest`.
then put high-level things like `SPVProof` and `BitcoinHeader` into a new `types.go`

That's still not ideal tho

ErinHales:
I think I've figured out the best solution.  Golang allows two packages in the same folder if the second one is a test package (e.g. mypackage_test).  So I separated the btcspv test files into a `btcspv_test` package, but they remain in the same file so the coverage still works.  This also fixed the import cycle problem.

ErinHales:
Ready for another review when you are :)

prestwich:
Can you rebase without the `coverage.out` file

golang/btcspv/bitcoin_spv_test.go
```diff
@@ -10,651 +9,577 @@ import (
...
+	ExtractOutputAtIndex         []tutils.ExtractOutputAtIndexTC       `json:"extractOutputAtIndex"`
+	ExtractOutputAtIndexError    []tutils.ExtractOutputAtIndexError    `json:"extractOutputAtIndexError"`
+	ExtractTarget                []tutils.ExtractTargetTC              `json:"extractTarget"`
+	ExtractTimestamp             []tutils.ExtractTimestampTC           `json:"extractTimestamp"`
+	Hash256MerkleStep            []tutils.Hash256MerkleStepTC          `json:"hash256MerkleStep"`
+	VerifyHash256Merkle          []tutils.VerifyHash256MerkleTC        `json:"verifyHash256Merkle"`
+	RetargetAlgorithm            []tutils.RetargetAlgorithmTC          `json:"retargetAlgorithm"`
+	CalculateDifficulty          []tutils.CalculateDifficultyTC        `json:"calculateDifficulty"`
+	Prove                        []tutils.ProveTC                      `json:"prove"`
+	CalculateTxId                []tutils.CalculateTxIDTC              `json:"calculateTxId"`
```
prestwich: nit: can this have a capital `ID` at the end?

prestwich:
nit, and remove the coverage file and we seem good to go

golang/btcspv/bitcoin_spv_test.go
ErinHales: So, this is like this pretty much everywhere which is why I did it like this.  Should I go ahead and change this in all the languages?

prestwich: No, just in Go. I'm fine with it being lowercase elsewhere. but the go linter hates it

ErinHales: Got it, this is already fixed then.  Should be good to merge

PR #136, opened 2020-03-25
Overhaul golang testing

ErinHales:
Updates Golang tests to use strongly typed cases instead of messy type assertions.


prestwich:
Ok, so I like this, but it breaks the coverage tool. Can we move just the `_test_types.go` files into a new folder and leave the `_test.go` files in the same folder?

ErinHales:
> Ok, so I like this, but it breaks the coverage tool. Can we move just the `_test_types.go` files into a new folder and leave the `_test.go` files in the same folder?

So, I think you meant to leave this comment in the other PR.  This one has all the testing files inside the `btcspv` folder.  That's why I separated out the other PR, in case we ran into problems or didn't like the structure.

prestwich:
oops yeah, you're right. :rofl:

PR #148, opened 2020-04-07
removed be/le from js client, fixed testing setup

### LINKED ISSUE #146: Update `js/client` to match new JSON format
The client still uses BE/LE. Needs to be updated to have only LE

ErinHales:
Closes #146 

 - No tests written for this yet, going to make issue.

### DIFF
--- a/js/clients/lib/BcoinClient.js
+++ b/js/clients/lib/BcoinClient.js
@@ -93,7 +93,6 @@ class BcoinClient extends NodeClient {
       vout: txinfo.vout,
       locktime: txinfo.locktime,
       tx_id: txid,
-      tx_id_le: reverse(txid),
       index,
       confirming_header: header,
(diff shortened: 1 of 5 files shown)


js/clients/lib/BcoinClient.js
```diff
@@ -93,7 +93,6 @@ class BcoinClient extends NodeClient {
       vout: txinfo.vout,
       locktime: txinfo.locktime,
       tx_id: txid,
-      tx_id_le: reverse(txid),
```
prestwich, line 96: these now return the BE format, right? we want that changed to LE

ErinHales: Oh, yes, you're right

prestwich:
okay. I have locally verified the client is working and produces good proofs

PR #154, opened 2020-05-21
add travis ci

emaG3m:


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "c/deps/ckb-c-stdlib"]
 	path = c/deps/ckb-c-stdlib
-	url = git@github.com:nervosnetwork/ckb-c-stdlib.git
```
emaG3m, line 3: ssh was getting a permissioned denied for travis

PR #155, opened 2020-05-22
test: vector for maliciously long witness scripts

prestwich:
TODO: `extractOutputScriptLength` is busted. Remove or update to support varints


prestwich:
Merging this now, opening an issue for `extractOutputScriptLength`. Currently nothing uses it

PR #158, opened 2020-05-25
Btc view

prestwich:
Tracking PR for solidity@4.0.0 updates. 
- [X] First draft of `BTCView`
- [x] `BTCView` Tests
- [x] First draft of `SPVView` (TODO: better name)
- [x] `SPVView` Tests
- [x] Update Sighash and signature utils to use BTCView instead of BTCUtils
- [x] Sig tests
- [x] Remove BTCUtils + Delegates


prestwich:
@ErinHales this is ready for review

solidity/contracts/CheckBitcoinSigs.sol
```diff
@@ -19,33 +17,39 @@ library CheckBitcoinSigs {
...
+
+        bytes20 _digest;
+        if (_pubkey.length == 33) {_digest = _pubkeyView.hash160();}
+        else if (_pubkey.length == 64) {
+            // This is less memory efficient than it ought to be
+            // But we immediately deallocate it, so it's usually no issue
+            uint8 prefix = uint8(_pubkey[_pubkey.length - 1]) % 2 == 1 ? 3 : 2;
+            bytes memory _x = _pubkey.ref(0).prefix(32, 0).clone();
+            bytes memory _compressed = abi.encodePacked(prefix, _x);
+            _digest = _compressed.ref(0).hash160();
```
ErinHales, line 39: This seems like it could be simplified... Maybe:
```
uint8 prefix = uint8(_pubkey[pubkey.length - 1]) % 2 + 2;
bytes memory _x = _pubkeyView.prefix(32, 0);
```
Also, does _x need to be saved to memory?  Could we do something like:
```
bytes memory _compressed =  abi.encodePacked(prefix, _pubkey.safeSlice(0, 32)); // not sure if there is a safeSlice equivalent in Solidity...
_digest = compressed.ref(0).hash160();
```

prestwich: > This seems like it could be simplified... Maybe:

Good call

> Also, does _x need to be saved to memory 

I chose to do it this way because of some rough edges on solidity memory. Because prefix isn't in memory, we can't use `unsafeJoin`. We could drop into assembly and store the prefix, then x. but I'd prefer to avoid doing that if it can be avoided. This uses 32 bytes more memory than I'd prefer, but I think it's okay

PR #160, opened 2020-05-26
Added docs to ViewBTC.sol

ErinHales:


solidity/contracts/ViewBTC.sol
```diff
@@ -41,11 +41,20 @@ library BTCUtils {
...
+    /// @notice             requires `memView` to be of a specified type
+    /// @param memView      a 29-byte view with a 5-byte type
+    /// @param t            the expected type (e.g. BTCTypes.Outpoint, BTCTypes.TxIn, etc)
+    /// @return             passes if it is the correct type, errors if not
     modifier typeAssert(bytes29 memView, BTCTypes t) {
         require(memView.typeOf() == uint40(t), "type assertion failed");
         _;
     }
 
+    // TODO: what does this do?
```
prestwich, line 53: functionally the same as `parseVarInt` but without the cruft. Reads a `VarInt`

prestwich: It reads a compact int from the view at the specified index

solidity/contracts/ViewBTC.sol
```diff
@@ -61,6 +70,9 @@ library BTCUtils {
         number = uint64(payload);
     }
 
+    /// @notice         converts an integer from uint64 to uint8
```
prestwich, line 73: this gives the total length (in bytes) of a CompactInt-encoded number

solidity/contracts/ViewBTC.sol
```diff
@@ -73,34 +85,56 @@ library BTCUtils {
...
+    /// @notice             extracts the index as an integer from the outpoint
+    /// @param _outpoint    the outpoint
+    /// @return             the index
     function outpointIdx(bytes29 _outpoint) internal pure typeAssert(_outpoint, BTCTypes.Outpoint) returns (uint32) {
         return uint32(_outpoint.indexLEUint(32, 4));
     }
 
+    /// @notice          extracts the outpoint from an input
+    /// @param _input    the input
+    /// @return          the outpoint as a typed bytes29
```
prestwich, line 104: say memory view instead of `bytes29`

solidity/contracts/ViewBTC.sol
```diff
@@ -117,24 +151,41 @@ library BTCUtils {
         return _vin.slice(_offset, _len, uint40(BTCTypes.TxIn));
     }
 
+    // TODO: what does this do?
```
prestwich, line 154: Extracts the raw LE bytes of the output value

solidity/contracts/ViewBTC.sol
```diff
@@ -117,24 +151,41 @@ library BTCUtils {
...
     }
 
+    /// @notice         extracts the value from an output
+    /// @param _output  the output
+    /// @return         the value
     function value(bytes29 _output) internal pure typeAssert(_output, BTCTypes.TxOut) returns (uint64) {
         return uint64(_output.indexLEUint(0, 8));
     }
 
+    /// @notice             extracts the script pub key from an output
```
prestwich, line 169: `scriptPubkey`

solidity/contracts/ViewBTC.sol
```diff
@@ -117,24 +151,41 @@ library BTCUtils {
...
+    /// @param _output      the output
+    /// @return             the script pub key
     function scriptPubkey(bytes29 _output) internal pure typeAssert(_output, BTCTypes.TxOut) returns (bytes29) {
         uint64 scriptLength = indexCompactInt(_output, 8);
         return _output.slice(8, compactIntLength(scriptLength) + scriptLength, uint40(BTCTypes.ScriptPubkey));
     }
 
+    /// @notice             determines the output length
+    /// @param _outputs     the outputs
```
prestwich, line 178: a `vout` without its length prefix

solidity/contracts/ViewBTC.sol
```diff
@@ -73,34 +85,56 @@ library BTCUtils {
...
+    /// @param _input   the input
+    /// @return         the sequence as a typed bytes29
     function sequence(bytes29 _input) internal pure typeAssert(_input, BTCTypes.TxIn) returns (uint32) {
         uint64 scriptLength = indexCompactInt(_input, 36);
         uint256 scriptEnd = 36 + compactIntLength(scriptLength) + scriptLength;
         return uint32(_input.indexLEUint(scriptEnd, 4));
     }
 
+    /// @notice         determines the input length
+    /// @param _input   the input
```
prestwich, line 127: the `vin` without its length prefix

solidity/contracts/ViewBTC.sol
prestwich: determines the length of the first output in an array of outputs

solidity/contracts/ViewBTC.sol
prestwich: determines the length of the first input in an array of inputs

solidity/contracts/ViewBTC.sol
```diff
@@ -151,6 +202,9 @@ library BTCUtils {
         return _vout.slice(_offset, _len, uint40(BTCTypes.TxOut));
     }
 
+    /// @notice         extracts the Op Return Payload
```
prestwich, line 198: note that it may return null if not a valid Op Return output

solidity/contracts/ViewBTC.sol
```diff
@@ -160,6 +214,9 @@ library BTCUtils {
         return _spk.slice(3, _payloadLen, uint40(BTCTypes.OpReturnPayload));
     }
 
+    /// @notice         extracts the payload from a script pub key
+    /// @param _spk     the script pub key
+    /// @return         the payload
     function payload(bytes29 _spk) internal pure typeAssert(_spk, BTCTypes.ScriptPubkey) returns (bytes29) {
```
prestwich, line 213: note that it may return null if not a valid PKH, SH, WPKH, or WSH output

solidity/contracts/ViewBTC.sol
```diff
@@ -229,49 +292,76 @@ library BTCUtils {
...
+    /// @param _arr the header array
+    /// @return     the typed header array
     function tryAsHeaderArray(bytes29 _arr) internal pure typeAssert(_arr, BTCTypes.Unknown) returns (bytes29) {
         if (_arr.len() % 80 != 0) {
             return TypedMemView.nullView();
         }
         return _arr.castTo(uint40(BTCTypes.HeaderArray));
     }
 
+    /// @notice     verifies the merkle array and converts to a typed bytes29
```
prestwich, line 308: not that each `tryAs` function will return null in error cases

solidity/contracts/ViewBTC.sol
```diff
@@ -229,49 +292,76 @@ library BTCUtils {
...
     }
 
+    /// @notice         extracts the parent hash from the header
+    /// @param _header  the header
+    /// @return         the parent hash
     function parent(bytes29 _header) internal pure typeAssert(_header, BTCTypes.Header) returns (bytes32) {
         return _header.index(4, 32);
     }
 
+    /// @notice
```
prestwich: calculates the Proof of Work hash of the header

PR #164, opened 2020-05-27
Spv view

prestwich:
Reimplements ValidateSPV using typed mem views

### DIFF
--- a/solidity/contracts/ViewBTC.sol
+++ b/solidity/contracts/ViewBTC.sol
@@ -303,6 +303,12 @@ library ViewBTC {
         return _header.castTo(uint40(BTCTypes.Header));
     }
 
+    function indexHeaderArray(bytes29 _arr, uint256 index) internal pure typeAssert(_arr, BTCTypes.HeaderArray) returns (bytes29) {
+        require(_arr.len() > index.add(1).mul(80), "Header array read overrun");
+        return _arr.slice(index.mul(80), index.add(1).mul(80), uint40(BTCTypes.Header));
+    }
+
+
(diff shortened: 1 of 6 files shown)


solidity/contracts/ViewBTC.sol
```diff
@@ -365,10 +378,17 @@ library ViewBTC {
...
     /// @param _header  the header
     /// @return         the Proof of Work hash
-    function work(bytes29 _header) internal view typeAssert(_header, BTCTypes.Header) returns (bytes32) {
+    function workHash(bytes29 _header) internal view typeAssert(_header, BTCTypes.Header) returns (bytes32) {
         return _header.hash256();
     }
 
+    /// @notice         calculates the Proof of Work hash of the header, and converts to an integer
+    /// @param _header  the header
+    /// @return         the Proof of Work hash
```
ErinHales, line 387: A little nit-picky, but maybe `the Proof of Work hash as an integer` just to be clear?

solidity/contracts/ViewSPV.sol
```diff
@@ -0,0 +1,116 @@
...
+
+import {TypedMemView} from "./TypedMemView.sol";
+import {ViewBTC} from "./ViewBTC.sol";
+import {SafeMath} from "./SafeMath.sol";
+
+
+library ViewSPV {
+    using TypedMemView for bytes;
+    using TypedMemView for bytes29;
+    using ViewBTC for bytes29;
```
ErinHales, line 14: I'm not very familiar with Solidity, so I might not be understanding this correctly...  But do we need to use `TypedMemView` and `ViewBTC` for bytes29?

PR #165, opened 2020-05-28
WIP: `viewBTC` js tests

ErinHales:


testVectors.json
```diff
@@ -615,8 +737,8 @@
       "output": "0x0000000000000000000000000000000000000000"
     },
     {
-      "input": "0x000000000000000017a914000000000000000000000000000000000000000087",
-      "output": "0x0000000000000000000000000000000000000000"
+      "input": "0x4897070000000000160014a4333e5612ab1a1043b25755c89b16d55184a42f",
+      "output": "0xa4333e5612ab1a1043b25755c89b16d55184a42f"
     }
```
ErinHales, line 742: Delete.  We both added this test case in different spots, resulting in doubles

prestwich: my bad :roll_eyes:

PR #166, opened 2020-05-28
Adds some additional docs fixes

sr-gi:
May add some more if I find any, leaving it as a draft for now.

### DIFF
--- a/solidity/contracts/BTCUtils.sol
+++ b/solidity/contracts/BTCUtils.sol
@@ -209,8 +209,8 @@ library BTCUtils {
         return (_varIntDataLen, _scriptSigLen);
     }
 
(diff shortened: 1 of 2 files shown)


prestwich:
Will more commits be added to this, or is it ready for merging?

sr-gi:
I'd wait since I'm still checking some of the files, and I may add some more if I find any. Will move from draft to PR once ready.

sr-gi:
As a side note, `validateVin` and `validateVout` are placed under Witness TX, but they work for both legacy and segwit.

prestwich:
Really the distinction between Legacy and Witness is no longer useful, and should probably be removed

PR #168, opened 2020-05-30
tests: Tests and bugfixes for ViewSPV

prestwich:
Several bugfixes, and tests.

### DIFF
--- a/solidity/contracts/ViewSPV.sol
+++ b/solidity/contracts/ViewSPV.sol
@@ -66,8 +66,8 @@ library ViewSPV {
         bytes29 _vout,
         bytes4 _locktime
     ) internal view returns (bytes32) {
-        bytes29 _ins = _vin.tryAsVin().assertValid();
-        bytes29 _outs = _vout.tryAsVin().assertValid();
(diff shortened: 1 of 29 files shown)


solidity/contracts/ViewSPV.sol
```diff
@@ -66,8 +66,8 @@ library ViewSPV {
         bytes29 _vout,
         bytes4 _locktime
     ) internal view returns (bytes32) {
-        bytes29 _ins = _vin.tryAsVin().assertValid();
-        bytes29 _outs = _vout.tryAsVin().assertValid();
+        bytes29 _ins = _vin.tryAsVin().assertValid();  // do we really need to check these?
+        bytes29 _outs = _vout.tryAsVout().assertValid();  // do we need to check these?
```
ErinHales, line 70: I was playing around with this and am going to make a PR

ErinHales:
@prestwich Any reason there is not a test for `checkWork`?

prestwich:
The function signature changed a bit, so now we need to redo the test with headers instead of digests. You're welcome to tackle it if you want

PR #170, opened 2020-06-01
View spv type assertions

ErinHales:
- Moves type conversions and validations to `ViewSPVtest.sol`
 - Adds type assertions to `ViewSPV.sol`.

### DIFF
--- a/solidity/contracts/ViewSPV.sol
+++ b/solidity/contracts/ViewSPV.sol
@@ -81,7 +81,7 @@ library ViewSPV {
     /// @param _header      Header view
     /// @param _target      The target threshold
     /// @return             true if header work is valid, false otherwise
-    function checkWork(bytes29 _header, uint256 _target) internal view returns (bool) {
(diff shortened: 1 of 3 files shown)


solidity/contracts/ViewSPV.sol
```diff
@@ -96,11 +100,10 @@ library ViewSPV {
     /// @param _headers     Raw byte array of header chain
     /// @return             The total accumulated difficulty of the header chain, or an error code
     function checkChain(bytes29 _headers) internal view returns (uint256 _totalDifficulty) {
-        bytes29 _headerChain = _headers.tryAsHeaderArray().assertValid();
```
prestwich, line 99: any reason to remove without replacing? I guess `indexHeaderArray` will make this check later :thinking:

ErinHales: Oh, thanks!  I added the validation check in `ViewSPVtest.sol`, but forgot to add the type assertion here.

PR #172, opened 2020-06-01
refactor: improve non-minimal varint error

prestwich:


solidity/contracts/ViewBTC.sol
```diff
@@ -52,10 +52,11 @@ library ViewBTC {
 
     /// Revert with an error message re: non-minimal VarInts
     function revertNonMinimal(bytes29 ref) private pure returns (string memory) {
+        (, uint256 g) = encodeHex(ref.indexUint(0, ref.len()));
```
ErinHales, line 55: `TypedMemView.encodeHex`

ErinHales: Oh, actually change that whole line to: `(, uint256 g) = TypedMemView.encodeHex(ref.indexUint(0, uint8(ref.len())));`

PR #182, opened 2020-07-22
js: vendor bcoin block and tx primitives

tynes:


prestwich:
`+103,264` 

o.O

tynes:
> `+103,264`
> 
> o.O

padding them contributor stats

on a more serious note, i used a webpack like tool to bundle the files. i could try to dedupe everything in `vendor`

PR #189, opened 2020-08-12
fix: make the no-std version work

huwenchao:
The original version will not work since the dependencies use `std` by default.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
 [[package]]
 name = "autocfg"
-version = "0.1.6"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


prestwich:
Updates cause `cargo test --no-default-features` to fail due to lack of a global allocator

huwenchao:
```
[lib]
crate-type = ["cdylib", "rlib"]
```

What is the code above used for? 
`cargo test --no-default-features` works without that.
And I checked some other no-std crates, they do not have that.

- https://github.com/KokaKiwi/rust-hex/blob/master/Cargo.toml
- https://github.com/rust-num/num-bigint/blob/master/Cargo.toml

huwenchao:
And even `cargo build --no-default-features` fails, it works to use it as a no-std dependency.

This is an example: <https://github.com/huwenchao/bitcoin-spv-lib/blob/master/contracts/bitcoin-spv-lib/Cargo.toml#L10>

prestwich:
the `lib` specifies linkages. See the docs here: https://doc.rust-lang.org/reference/linkage.html

prestwich:
Check out #190 and let me know if it works for you

huwenchao:
> Check out #190 and let me know if it works for you

It works! Thanks a lot.

